PR #105, opened 2019-03-08
Adds descriptive metadata part 2

### LINKED ISSUE #86: Add descriptive MD to curate-generic worktype, pt 2
- [ ] Add descriptive MD fields 11-20, following notes from #61
- [ ] Test for presence of MD
- [ ] Test number of accepted terms (multiple: true/false)
- [ ] Test indexing in Solr (searchable, facetable, etc.)
- [ ] Test for requirement (required, optional, etc.)

devanshu-m:
1. This commit adds 10 more (10-20) descriptive metadata.
2. Comments out inclusion of Hyrax's Basic metadata - We may not need all of the basic and core metadata. Commenting out the inclusion will allow us to define our required basic and core metadata.
3. Adds `publisher` and `date_created` from basic metadata to forms.
4. Edits `create_curate_generic_work_spec` to match available metadata properties - since we do not have basic and core metadata available, we will need to write tests with our set of metadata instead of tests with default metadata properties.

Closes #86

### DIFF
--- a/app/forms/hyrax/curate_generic_work_form.rb
+++ b/app/forms/hyrax/curate_generic_work_form.rb
@@ -4,6 +4,6 @@ module Hyrax
   # Generated form for CurateGenericWork
   class CurateGenericWorkForm < Hyrax::Forms::WorkForm
     self.model_class = ::CurateGenericWork
-    self.terms += [:resource_type]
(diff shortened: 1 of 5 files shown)


alexBLR:
👍

Issue #110, opened 2019-03-14
Test SolrDoc for new metadata


rotated8:
1. Create/Save a work with a specific set of metadata
1. Retrieve the work from Solr (there is a Solr query service for this, you should get a Solr document back)
1. Test that the stored metadata is stored into Solr correctly.

The metadata should span the matrix of types of MD we put into Solr (single/multiple, faceted, searchable, etc.)

This might be helpful: https://github.com/samvera/hydra-head/wiki/Solr-Schema

rotated8:
The objective of these tests is to ensure the metadata we have in the generic work are stored in Solr in the way we expect. To that end, we are interested in ensuring the metadata is present in the Solr doc, and that the field that it is stored in matches our expectations.

PR #111, opened 2019-03-15
Adds initial work for testing SolrDoc with new metadata

devanshu-m:
This is an `in progress` PR.

I have included BasicMetadata so that a work could be created without errors. This will cause some tests to fail (for example: we have tests written for `publisher` to be single-valued, however it is multi-valued in Hyrax::BasicMetadata).

Unit tests written for checking metadata returned by SolrDoc also are failing right now. I am not sure why `primary_language` and `abstract` have suffix `_tesim` when they should have `_tesi` and `_tes` respectively. This is something we might need to look into.

Connected to #110 

Update (04/08): The final commit tests if SolrDoc saves and returns the saved values correctly. This test will currently fail because of the exclusion of `BasicMetadata`


devanshu-m:
image

Failing test

alexBLR:
Fixed failing tests. It should be good to go now.

PR #117, opened 2019-03-19
Update capistrano

SolomonHD:
This commit should allow for the developers to deploy to test and production from their development machines.


config/deploy.rb
```diff
@@ -14,6 +14,12 @@
 append :linked_dirs, "log", "public/assets", "tmp/pids", "tmp/cache", "tmp/sockets", "config/emory"
 append :linked_files, ".env.production", "config/secrets.yml"
 
+set :default_env, {
+        'PASSENGER_INSTANCE_REGISTRY_DIR' => '/var/run',
+	'PATH' => '$PATH:/opt/rh/rh-ruby25/root/usr/local/bin:/opt/rh/rh-ruby25/root/usr/bin',
```
rotated8: You seemed unhappy about having to hardcode this. Do we need to open a ticket to revisit it in the future?

SolomonHD: I would love to have an environment variable here instead, would be more flexible in case we ever needed to switch ruby versions and it should be possible in capistrano but I can't figure it out. I'll revisit this later.

config/deploy.rb
```diff
@@ -1,11 +1,11 @@
 # config valid for current version and patch releases of Capistrano
 lock "~> 3.11.0"
 
+
```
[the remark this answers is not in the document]
SolomonHD: I just pushed up a fix for this

config/deploy.rb
rotated8: That sounds like a yes.

config/deploy.rb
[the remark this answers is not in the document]
rotated8: Sorry, I was talking about both lines 3 and 4 being blank.

alexBLR:
👍

PR #129, opened 2019-03-28
Validates fileset use metadata against controlled vocab

### LINKED ISSUE #83: Validate fileset use metadata against controlled vocab
- [x] Validate FileSet.pcdm_use so it is set to only one of the terms in the controlled vocab from #82
- [x] Test that validation happens correctly

alexBLR:
closes #83


alexBLR:
yes. it is covered in the other ticket

Issue #132, opened 2019-04-01
Use Shibboleth on curate-test


rotated8:
- [x] Change env variable to use Shibboleth.
- [x] Test that login works
- [x] Ensure the admin setup rake task is run.
- [x] Revert changes that were made in order to disable database_auth on curate-test

tmiles2:
@devanshu-m what part of this work has been completed? I don't see any tasks checked off but there is a PR.

devanshu-m:
@tmiles2 I will add a task to this list now.

PR #149, opened 2019-04-04
Edits to make prod shib ready

devanshu-m:
This commit disables database_auth on prod and makes it shib ready. Database_auth will only be available on dev now on.

This commit also updates gem versions.

Connected to #132


alexBLR:
looks great!

PR #164, opened 2019-04-16
Removes Hyrax's basic metadata

rotated8:
Removes the concern from our generic work type, and stops trying to index those terms.


alexBLR:
Looks. great!

PR #165, opened 2019-04-16
Makes factory traits dynamic

rotated8:
Rubocop complains that the current factories use static traits, which will be removed in a upcoming version of FactoryBot. This changes them to be dynamic, and removes those warnings.


alexBLR:
👍

PR #166, opened 2019-04-16
Updates Rubocop and adds Pry for debugging.

rotated8:
- Push Rubocop to use Ruby 2.5 syntax. This pushes the admins task, and turning off the frozen string literal changes.

- Enable the ABC Size metric, and limit the others that are still off to only the failing files.

- The User model had two problems. First, the ABC size was too high, and second, AR's #first_or_create method has been silently unsupported. Switching to #find_or_initialize_by solves the second problem, and
using #assign_attributes fixes the first.

- Pry can do debugging like byebug, but adds a bunch of convenience. Use `binding.pry` where you would have used `byebug`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,14 +1,31 @@
 inherit_gem:
   bixby: bixby_default.yml
 
-Metrics/AbcSize:
-  Enabled: false
+AllCops:
+    TargetRubyVersion: 2.5
+    Exclude:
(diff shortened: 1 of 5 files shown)


rotated8:
Since this changes the way we save users, this will need testing after it is merged and deployed. (Attn: @eporter23 @SolomonHD)

alexBLR:
🌮

PR #196, opened 2019-05-06
WIP Adds webpacker and yarn

rotated8:
Webpacker and yarn will replace the old Rails asset pipeline, and this gets us ahead of the curve. It also updates Hyrax to Beta 2.


rotated8:
This branch is built off the Circle one, so it should not be merged until it has been rebased.

PR #213, opened 2019-05-17
Adds ability to attach multiple files to a single fileset

devanshu-m:
This PR uncomments failing tests in `attach_files_to_work_job_spec.rb`


rotated8:
This should not be squashed when it is merged. There are three separate pieces of development here I'd like to keep.

PR #244, opened 2019-06-07
Fix date validation

rotated8:
Dates are validated with javascript before the form is saved. In order to preserve EDTF dates, they should not be cast to the Ruby date type.

### DIFF
--- a/app/models/curate_generic_work.rb
+++ b/app/models/curate_generic_work.rb
@@ -1,5 +1,4 @@
-# Generated via
-#  `rails generate hyrax:work CurateGenericWork`
+# frozen_string_literal: true
 class CurateGenericWork < ActiveFedora::Base
   include ::Hyrax::WorkBehavior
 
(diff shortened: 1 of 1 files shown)


app/models/curate_generic_work.rb
```diff
@@ -1,13 +1,11 @@
...
-#  `rails generate hyrax:work CurateGenericWork`
+# frozen_string_literal: true
 class CurateGenericWork < ActiveFedora::Base
   include ::Hyrax::WorkBehavior
 
   self.indexer = CurateGenericWorkIndexer
   # Change this to restrict which works can be added as a child.
   # self.valid_child_concerns = []
   validates :title, presence: { message: 'Your work must have a title.' }
-  validates :date_created, :date_issued, :conference_dates, :copyright_date, :scheduled_rights_review, type: Date
```
rotated8: By not enforcing a `Date` type here, we may need to change how we index this data.

PR #256, opened 2019-06-13
Initial work to edit form to add multiple files to a single fileset

devanshu-m:


devanshu-m:
@alexBLR @rotated8 This should be ready to go now. I have made changes as per suggestions from today morning.

PR #277, opened 2019-07-01
Adds collection metadata properties

devanshu-m:
* Adds metadata properties to collection model
* Edits Hyrax's collection form to add new collection md properties
* Edits Hyrax's collection form tests to add new md properties and adds collections factory


app/forms/hyrax/forms/collection_form.rb
```diff
@@ -0,0 +1,156 @@
+# [Hyrax-overwrite]
+module Hyrax
+  module Forms
+    # rubocop:disable Metrics/ClassLength
```
rotated8: Please move this to the Rubocop overrides file.

PR #302, opened 2019-07-17
Fixes a warning from the Passenger logs

rotated8:
WARN -- : Cannot enable realtime notifications atop Passenger + Apache.
Coercing `Hyrax.config.realtime_notifications` to `false`. Set this
value to `false` in config/initializers/hyrax.rb to stop seeing this
warning.

### DIFF
--- a/config/initializers/hyrax.rb
+++ b/config/initializers/hyrax.rb
@@ -102,7 +102,7 @@ Hyrax.config do |config|
   # config.arkivo_api = false
 
   # Stream realtime notifications to users in the browser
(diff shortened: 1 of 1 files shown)


rotated8:
The only change you should see for this is when you restart Apache, the comment from above no longer shows up.

mprefer:
Changed config.realtime_notifications=true to
               config.realtime_notifications=false
in /opt/dlp-curate/current/config/initializers/hyrax.rb on both curate and curate-test.
Performed 'service httpd reload' to reload Apache + Passenger.

PR #304, opened 2019-07-18
Testing visibility

alexBLR:


spec/features/create_curate_generic_work_spec.rb
```diff
@@ -110,6 +110,58 @@
...
+      expect(cgw.visibility).to eq 'restricted'
+
+      visit("/concern/curate_generic_works/#{cgw.id}/edit")
+
+      find('body').click
+      choose('curate_generic_work_visibility_authenticated')
+      click_on('Save')
+
+      cgw.reload
+      expect(cgw.visibility).to eq 'authenticated'
```
rotated8: If we have a block here when the user logs out and attempts to visit the page, that will show that the visibility is being enforced.

spec/features/create_curate_generic_work_spec.rb
```diff
@@ -110,6 +110,58 @@
...
+      expect(cgw.visibility).to eq 'restricted'
+
+      visit("/concern/curate_generic_works/#{cgw.id}/edit")
+
+      find('body').click
+      choose('curate_generic_work_visibility_restricted')
+      click_on('Save')
+
+      cgw.reload
+      expect(cgw.visibility).to eq 'restricted'
```
rotated8: Similar to the other comment, but with a different user logged in.

PR #341, opened 2019-07-27
Initial setup for zizia importer

bess:
* Add a very basic zizia import
* Add a very basic metadata mapping (only title for right now)
* Cleanup between test runs for better test reliability
* Add a very basic CSV fixture object -- this will grow as we know more
about the data we're trying to import

Connected to https://github.com/curationexperts/dlp-curate/issues/11


bess:
curate_basic_command_line_import

PR #369, opened 2019-08-01
Add honeybadger for error logging

bess:
Using an error tracking system like honeybadger makes it much easier to
know about errors in the production environment, to record enough
information that they can be easily diagnosed, and to track whether they
re-occur.

See https://app.honeybadger.io/ for more info.


bess:
honeybadger_curate_cd

PR #371, opened 2019-08-01
Change Rights Statement Properties

rotated8:
A controlled vocab for rights statement property is available from Hyrax, so we use that. For our free text rights statement, change the property name, and move the field on the form to conform with the metadata spreadsheet.


app/models/curate_generic_work.rb
```diff
@@ -181,9 +181,9 @@ def assign_id
     index.as :stored_searchable
   end
 
-  property :rights_statement, predicate: "http://purl.org/dc/elements/1.1/rights"
+  property :rights_statement_text, predicate: "http://purl.org/dc/elements/1.1/rights"
 
-  property :rights_statement_controlled, predicate: "http://www.europeana.eu/schemas/edm/rights", multiple: false do |index|
+  property :rights_statement, predicate: "http://www.europeana.eu/schemas/edm/rights" do |index|
```
rotated8: @bess @little9 Y'all might want to see this part.

rotated8:
@little9 @bess I ran the load_sample_data rake task, and did not get any errors, but y'all may want to take a look.

PR #424, opened 2019-08-07
Update CircleCI to use latest Samvera orb

mark-dce:
The orb release being referenced wasn't a valid release number according to https://circleci.com/orbs/registry/orb/samvera/circleci-orb.  Trying the latest samvera/circleci-orb version to see if it fixes coveralls notifications.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-    samvera: samvera/circleci-orb@dev:fix-bundle-cache
+    samvera: samvera/circleci-orb@0.3.1
 jobs:
     build:
         parameters:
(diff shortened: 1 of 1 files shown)


rotated8:
We were using an dev version because we needed [this PR](https://github.com/samvera-labs/samvera-circleci-orb/pull/16/), which wasn't available at the time.

I think this needs the [parallel build webhook](https://docs.coveralls.io/parallel-build-webhook) as well.

mark-dce:
Totally agreed about the webhook...

I've been reading through the samera repos (and their circle CI projects) that use the same orb and they don't show the parallel build web hook anywhere which is why I'm hypothesizing it's burned somewhere in the orb code. 

I'll ask my friends in the Component Maintenance WG to see if any of them have run across this and how to fix it.

PR #432, opened 2019-08-08
Testing Coveralls

rotated8:


rotated8:
Circle [seems to think](https://circleci.com/docs/2.0/code-coverage/#coveralls) nothing special needs to be done for Coveralls...

little9:
@rotated8 I was having problems with their docs as well. Maybe try codecov.io? It should be very similar now that you've added the SimpleCov specific bits already.

PR #435, opened 2019-08-09
Add UV config

little9:
This adds the basic UV config to turn it on
and have it display images in developement
mode. If the `IIIF_SERVER_URL` is not set
it will use `Riiif` to display the images.

When Cantaloupe is setup that variable should
be set.


little9:
Screenshot from 2019-08-09 12-10-32

Issue #394, opened 2019-08-06
Automate creation of "Library Collection" Type


tmiles2:
See epic #383

Collection Type [settings documented here](https://docs.google.com/document/d/1w0iRk6TfawMWTKxiqHBUWpB5rRtqzcerNl55t5IYplg/edit#).

- [x] Document Collection Type settings for _Library Collection_ from Emily's manual configuration 
- [x] Automatically recreate this Collection Type when application is rebuilt

PR #438, opened 2019-08-09
Curate collection type participants

bess:
Connected to https://github.com/emory-libraries/dlp-curate/issues/394


bess:
Screen shot of auto-created CollectionType, on the participant's tab. I also tested this against curate-qa, which already had a "Library Collection" collection object with the wrong participant set, and it reconfigured the existing object as expected. 

library_collection_participants

PR #459, opened 2019-08-13
Adds ability to download files in a fileset

devanshu-m:
- Changes original_file symbol to preservation_master_file for default_content
- Adds `use` param to link_to for additional_files that should be passed along with filesetID


SolomonHD:
@devanshu-m It looks like circleci is failing, is this fixtures path being set anywhere in /config/initializers/hyrax.rb? That's where all the other paths are being set (derivates path, working path, etc)

devanshu-m:
@SolomonHD I don't think this failure was related to that. CI couldn't find the file because I had renamed it in fixtures, but forgot to rename it in the test. It's green now.

Issue #387, opened 2019-08-05
View Collection-level Metadata for an Emory Library Collection


eporter23:
## Story
As a Collection Manager or Metadata Specialist, I want to preview Descriptive, Rights, or Administrative metadata added to a Collection entity, so that I can spot-check its accuracy during migration testing

## Acceptance Criteria
_Use one or more of the following options to provide acceptance criteria._

#### Notes

For August sprints, enable viewing of all populated metadata on the Collection View page. We can fine-tune which fields display to which users at a later time.

[Metadata Implementation - Collections](https://docs.google.com/spreadsheets/d/17WLINPefxb0cQh2Jgx6H7XnUCDH8Ol1WlKaxtBAGW38/edit#gid=1472024349)

We can use the same approach/layout conventions as are implemented for previewing work metadata.

Checklist:
- [x] All populated fields display (omit empty fields)
- [x] Fields display using the Form Field Order in the spreadsheet above
- [x] Fields use Display Labels from the spreadsheet above

bess:
![collection_metadata.png](https://images.zenhubusercontent.com/5ca3a84155cfe32f2f66b74c/d4799e71-d736-48b3-951c-5433cc388f5b)

PR #460, opened 2019-08-13
Display collection metadata

bess:
Connected to https://github.com/emory-libraries/dlp-curate/issues/387


bess:
collection_metadata

PR #514, opened 2019-08-22
Add a downloadable CSV to the metadata guide

little9:
This provides a link to a CSV representation
of the information displayed in the guide.


little9:
This isn't floated to the right. I think it will be easier to discover if it is on the left.

image

PR #530, opened 2019-08-28
implements preservation workflow model

### LINKED ISSUE #499: Time boxed preservation workflow implementation
- [x] Create preservation workflow model
- [x] Add curate work property
- [x] Write functional and unit tests
- [x] Assess approach after 1 day of implementation

alexBLR:
Closes #499


app/actors/hyrax/actors/curate_generic_work_actor.rb
```diff
@@ -3,9 +3,16 @@
...
-        # byebug
+        super
+        # ActiveFedora fails to propgate changes to nested attributes to
+        # `#resource` when indexed nested attributes are used. We force the
+        # issue here to work around for form edits.
+        (env.attributes.keys && KNOWN_NESTED_ATTRIBUTES).each do |attribute_key|
+          attribute = attribute_key.to_s.gsub('_attributes', '').to_sym
+          env.curation_concern.send(attribute).each { |member| member.try(:persist!) }
+        end
         super
```
rotated8: This `super` is not needed.

app/models/preservation_workflow.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require 'active_triples'
+
+class PreservationWorkflow < ActiveTriples::Resource
+  include ActiveTriples::RDFSource
+
+  property :workflow_type, predicate: "http://metadata.emory.edu/vocab/cor-terms#workflowType"
```
rotated8: Please check if we need to ad `multiple: false` for these properties.

spec/models/curate_generic_work_spec.rb
```diff
@@ -19,6 +19,10 @@
       w_members = work1.members
       expect(w_members.map(&:id)).to match_array [work2.id, work3.id, fileset1.id, fileset2.id]
     end
+
+    it "has a preservation workflow which is a PreservationWorkflow object" do
+      expect(work1.preservation_workflow.build).to be_instance_of PreservationWorkflow
+    end
```
rotated8: We should add a test that adds a workflow and saves the work. The test would then show that the property includes the new workflow.

PR #537, opened 2019-08-29
Edits FileActor to run characterizejob only on pmf [WIP]

devanshu-m:


rotated8:
Please let me know if the FITS output shows up in your local tests, but not in CI. If that is the case, we can install FITS in CI, which should solve the issue.

devanshu-m:
FITS is not installed in CI. I tried to find a way to mock the CharacterizeJob, however, was unsuccessful.

PR #538, opened 2019-08-29
Rename properties

rotated8:
* Properties are plural if multiple: true
Exceptions: title, creator- because of Hyrax; preservation_workflow- because it's in development

* Properties should be alphabetized in translations, work model, and collection model.

* Labels should always be singular, and have the form `property_name (Display Label)`


rotated8:
@bess I think I got everything, I can take or leave the last commit.

rotated8:
@bess @mark-dce If this is ok, I can put in a PR for the CSV headers, too.

bess:
@rotated8 This looks great. Please do put in a PR for the CSV headers too.

PR #557, opened 2019-09-03
Adds use to fileset show page

devanshu-m:


app/models/solr_document.rb
```diff
@@ -25,4 +25,8 @@ class SolrDocument
 
   use_extension(Hydra::ContentNegotiation)
   include SolrDocumentAccessors
+
+  def pcdm_use
+    self[Solrizer.solr_name('pcdm_use')]
```
rotated8: Please change this to use the field name explicitly, as Solrizer is being deprecated.

PR #559, opened 2019-09-04
Moves the admins yaml to a new groups dir in the Emory config.

rotated8:


SolomonHD:
I added one small change to capistrano so it will create config/emory/groups as the linked dir instead of config/emory

PR #565, opened 2019-09-04
Add renderers and label services for rights_statement and content_type

little9:
image


little9:
Connected to curationexperts/in-house#254
Connected to curationexperts/in-house#253

PR #571, opened 2019-09-06
Uses the FileSet to control the file use vocabulary.

rotated8:


rotated8:
@bess @little9 👀 FYI

rotated8:
@devanshu-m I think we can do that for now, and if it needs to change later, we can change it at that point.

PR #583, opened 2019-09-10
Adds ability to create user_groups and add users to those groups

devanshu-m:


lib/user_setup.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  # Add user to specific group
+  # @param [String] the uid of the user
+  # @return [User] the user who was just added to the group
+  def add_user_to_group(uid, provider = "database")
+    @logger.debug "Adding user #{uid} to #{@user_group}"
+    user = ::User.find_or_create_by(uid: uid)
+    user.password = "123456" if set_default_password?
```
alexBLR: If the user already exists, we want to make sure that we don't reassign values to the existing user.

db/seeds.rb
```diff
@@ -10,3 +10,11 @@
 a = AdminSetup.new
 # Setup admins
 a.setup
+
+require 'user_setup'
+Dir.glob("#{::Rails.root}/config/emory/groups/user_groups/*.yml") do |yml_file|
+  # Initialize UserSetup class
+  u = UserSetup.new(yml_file)
+  # Add users to user_group
```
alexBLR: it should be a part of the group folder.

PR #594, opened 2019-09-17
Adds user-groups to langmuir collection

devanshu-m:


app/importers/collection_permission_ensurer.rb
```diff
@@ -1,41 +1,50 @@
 # frozen_string_literal: true
 
 class CollectionPermissionEnsurer
-  AGENT_ID = 'admin'
+  # AGENT_ID = 'admin'
```
rotated8: Rather than commenting this out, please remove it.

app/importers/collection_permission_ensurer.rb
```diff
@@ -1,41 +1,50 @@
...
       hpt = find_or_create_permission_template(collection)
-      next if repo_admins_have_manage_rights?(hyrax_permission_template: hpt, access_permission: access_permission)
-      hpta = Hyrax::PermissionTemplateAccess.new
-      hpta.permission_template_id = hpt.id
-      hpta.agent_type = AGENT_TYPE
-      hpta.agent_id = AGENT_ID
-      hpta.access = access_permission
-      hpta.save
+      groups.each do |group|
+        next unless Role.exists?(name: group)
```
rotated8: Try filtering the groups list down before this loop. I think you can call `map(something)` on the groups and get only the groups that exist back. (or chain `compact` as well).

app/importers/collection_permission_ensurer.rb
```diff
@@ -1,41 +1,50 @@
...
-        existing_hpta = Hyrax::PermissionTemplateAccess.where(
-          permission_template_id: hyrax_permission_template.id,
-          agent_type: AGENT_TYPE,
-          agent_id: AGENT_ID,
-          access: access_permission
-        ).count
-        return true if existing_hpta.positive?
-        false
-      end
+    def repo_admins_have_manage_rights?(hyrax_permission_template:, access_permission:, group:)
```
rotated8: Let's change this to `group_rights_exist?`

PR #595, opened 2019-09-18
Add custom access levels to Zizia mapper

little9:


spec/system/import_langmuir_from_csv_spec.rb
```diff
@@ -46,18 +46,26 @@
       expect(page).to have_content 'Testing Collection'
 
       # Let the background jobs run, and check that the expected number of records got created.
-      expect(CurateGenericWork.count).to be 5
+      expect(CurateGenericWork.count).to eq 9
```
bess: Hey, why did the counts change? Mostly this looks great, but I'm confused about this part. I'd expect the importer to still attach the front and back of a photo to the same object, which for the 10 line fixture object means 5 objects, not 9.

little9: I saved some bad changes to the csv fixture. This is back to normal now.;

PR #599, opened 2019-09-20
displays preservation workflow on show page

### LINKED ISSUE #527: Display Preservation Workflow Fields

alexBLR:
Closes #527


app/controllers/hyrax/curate_generic_works_controller.rb
```diff
@@ -9,11 +9,17 @@ class CurateGenericWorksController < ApplicationController
 
     # Use this line if you want to use a custom presenter
     self.show_presenter = Hyrax::CurateGenericWorkPresenter
-
```
rotated8: Let's keep this whitespace, since it comes right before a method declaration.

app/models/curate_generic_work.rb
```diff
@@ -14,6 +14,13 @@ def assign_id
   validates :related_publications, url: true, if: -> { related_publications.present? }
   validates :related_datasets, url: true, if: -> { related_datasets.present? }
   validates :rights_documentation, url: true, if: -> { rights_documentation.present? }
+  before_save :index_preservation_workflow_terms
+
+  property :preservation_workflow, predicate: "http://metadata.emory.edu/vocab/cor-terms#preservation_workflow", class_name: "PreservationWorkflow"
+
+  property :preservation_workflow_terms, predicate: "http://metadata.emory.edu/vocab/cor-terms#preservation_workflow_attributes" do |index|
```
rotated8: The properties are alphabetized now, so these should move down to the 'P's.

app/models/curate_generic_work.rb
```diff
@@ -269,6 +276,10 @@ def assign_id
     index.as :stored_searchable
   end
 
+  def index_preservation_workflow_terms
```
rotated8: I think this ended up in the properties during a rebase. Please move after `volume`.

app/controllers/hyrax/curate_generic_works_controller.rb
alexBLR: fixed

app/models/curate_generic_work.rb
alexBLR: fixed

app/models/curate_generic_work.rb
alexBLR: done

PR #600, opened 2019-09-20
Remove visibility constant when loading Rails so it can be re-defined

little9:
This constant needs to be removed from `hydra-access-controls` and
replaced with out custom list. In prod, it is not enough to re-define
it in an initializer. It needs to be removed explicility.


little9:
image

bess:
Good catch! I will approve as soon as CI is green.

PR #649, opened 2019-10-03
Cleans up curate_collections rake task and CSVs

devanshu-m:


spec/system/migration_collections_spec.rb
```diff
@@ -4,13 +4,15 @@
...
     Curate::CollectionType.find_or_create_library_collection_type
-    CurateCollectionImporter.new.import(langmuir_csv_file)
-    Collection.last
+    CurateCollectionImporter.new.import(collections_csv_file)
   end
+  let(:langmuir) { Collection.where(local_call_number: "MSS1218").first }
+  let(:yellowbacks) { Collection.where(local_call_number: "YELLOWBACKS").first }
 
   context 'logged in as an admin user' do
     before do
```
rotated8: Can this be removed?

spec/importers/curate_collection_importer_spec.rb
```diff
@@ -3,22 +3,21 @@
...
   end
 
   it 'only makes one collection object per call number' do
-    cci.import(langmuir_csv)
-    expect(Collection.count).to eq 1
-    cci.import(langmuir_csv)
-    expect(Collection.count).to eq 1
+    cci.import(collections_csv)
+    expect(Collection.where(local_call_number: "MSS1218").count).to eq 1
+    expect(Collection.where(local_call_number: "YELLOWBACKS").count).to eq 1
```
rotated8: Try calling `cci.import` again after this line, and testing that the counts stay the same.

lib/tasks/curate_collections.rake
```diff
@@ -3,33 +3,9 @@ namespace :curate do
...
-      Rake::Task["curate:collections:yearbooks_setup"].invoke
-      Rake::Task["curate:collections:yellowbacks_setup"].invoke
-    end
-
-    desc 'Langmuir collection setup'
-    task langmuir_setup: [:environment] do
-      collection_pre_reqs
-      langmuir_csv_file = Rails.root.join('config', 'collection_metadata', 'langmuir_collection.csv')
-      CurateCollectionImporter.new.import(langmuir_csv_file)
-      puts "Langmuir collection object created"
```
rotated8: Can we add a `puts` to the importer so we still get output for each collection as it is created?

PR #670, opened 2019-10-09
Add update logic to ModularImporter

little9:
This continues the use of the decorator pattern
to add the logic for the update options to the
importer.

The import_langmuir_from_csv spec has been updated
to test for each of the new conditions.


app/importers/modular_importer.rb
```diff
@@ -79,4 +83,10 @@ def attach_files
     open_preservation_master_file.close
     open_intermediate_file.close
   end
+
+  def existing_work
+    CurateGenericWork.all.select do |c|
+      c.deduplication_key == @mapper.metadata['deduplication_key']
+    end
```
mark-dce: maybe call `.first` on line 90 instead of having to repeatedly call it in "app/importers/modular_importer_update_decorator.rb"

little9: I'd change this, but there are part of the code where it needs to behave like an `Array`.

This method has been changed to use `where` instead of `all.select`

PR #672, opened 2019-10-10
Add collection membership

little9:
This adds imported works to the selected collection.

### DIFF
--- a/app/importers/modular_importer.rb
+++ b/app/importers/modular_importer.rb
@@ -81,10 +81,19 @@ class ModularImporter
 
   def attach_files
     @uploaded_file = create_hyrax_uploaded_file
+    add_to_collection
     AttachFilesToWorkJob.perform_now(CurateGenericWork.find(work_id), [uploaded_file])
-    CurateGenericWork.find(work_id).reload
     open_preservation_master_file.close
     open_intermediate_file.close
+    CurateGenericWork.find(work_id).save
+  end
+
+  def add_to_collection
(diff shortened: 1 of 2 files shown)


app/importers/modular_importer.rb
```diff
@@ -81,10 +81,20 @@ def create_hyrax_uploaded_file
...
+    CurateGenericWork.find(work_id).save
+  end
+
+  def add_to_collection
+    return unless collection_id
+    collection = Collection.find(collection_id)
+    work = CurateGenericWork.find(work_id)
+    work.member_of_collections << collection
+    collection.save
+    work.save
```
mark-dce: Can you confirm the work.save is really necessary in line 97?  I thought we only had to save the collection.

spec/system/import_langmuir_from_csv_spec.rb
```diff
@@ -85,6 +85,7 @@ def default_update(page)
       visit "/dashboard/works"
       click_on work.title.first
       expect(page).to have_content work.title.first
+      expect(work.member_of_collections).not_to eq []
```
mark-dce: Would it be better to match the specific collection you're expecting in case some other test doesn't clean up properly?

app/importers/modular_importer.rb
mark-dce: Can you try this without line 97 and see if it works?  (if not everything else looks good and I'll approve)

PR #709, opened 2019-10-29
Modifies CSV as per requirements

devanshu-m:
* This commit modifies CSV for collections generation as per
requirements.
* Fixture for the same has also been modified.


config/collection_metadata/collections.csv
```diff
@@ -1,22 +1,6 @@
...
+Emory University Yearbooks,Emory Universit,"Stuart A. Rose Manuscript, Archives, and Rare Book Library | Oxford College Library (Oxford, Ga.)",Emory University Archives,"Rose Library Woodruff Library Emory University 540 Asbury Circle Atlanta, GA 30322 rose.library@emory.edu 404-727-6887","Emory student yearbooks including Oxford, Emory College, and Emory University.",English,,,,Emory University--Periodicals,,,Emory University,,,,,,,,,,,admin | repository_manager | rose_manager | oxford_manager,,rose_viewer | oxford_viewer
```
eporter23: Please add a "y" to _Emory_ Universit_ for institution on the Yearbooks collection on line 3

rotated8: Please add the deduplication key from #707 as well. For collections that do not have a call number, we can use title for the moment.

rotated8:
Please add the deduplication key from #707 as well. For collections that do not have a call number, we can use title for the moment.

PR #719, opened 2019-11-04
Adds FileSet use category edit field to FileSetEditForm

devanshu-m:
* This commit changes the form_class for FileSet form
to now use our new form_class, Curate::Forms::FileSetEditForm
which inherits the Hyrax::Forms::FileSetEditForm class, therefore,
we get everything from the hyrax class but we also add our own terms
for Curate in the Curate::Forms::FileSetEditForm class.
* Also, removes existing spec for Hyrax's fileset edit form and
replaces an identical spec for Curate's fileset edit form with new
terms.
* Edits the edit_file system spec to make sure the FileSet use
category can actually be edited from the FileSet edit page.


devanshu-m:
Before:
image

After:
image

PR #726, opened 2019-11-06
Create PreIngestWorks only for rows with the correct type

little9:
Before this commit `PreIngestWork`s were being created for
every row in the CSV. This adds logic to only create the
`PreIngestWork` for rows that have the value `work` in the
`type` column. `PreIngestFile`s are then attached to the
correct `PreIngestWork` based on the `deduplication_key`.

After this commit the batch status page will be correctly
grouped.


little9:
image

PR #749, opened 2019-11-12
Updates where work related preservation_events occur

devanshu-m:
* This commit overrides the `create` method from Hyrax::Actors::BaseActor
and adds three preservation_events that occur immediately after
work creation.
* Also, removes create_work_job and its spec. We don't need the
modifications in those files anymore since the create_work_job is
used in the batch_create_job.
* Edits the preservation event create method so that now we save without
the bang in the end. This will prevent from trying to run validations
again.


spec/actors/hyrax/actors/curate_generic_work_actor_spec.rb
```diff
@@ -1,9 +1,55 @@
...
+    let(:uploaded_file) { Hyrax::UploadedFile.create(file: file, user: user) }
+    let(:terminator) { Hyrax::Actors::Terminator.new }
+
+    before do
+      allow(terminator).to receive(:create).and_return(true)
+    end
+
+    context 'success' do
+      before do
+        redlock_client_stub
```
rotated8: Let's try the test without this. Circle should have Redis, so I don't think we need to stub the connection.

app/lib/preservation_events.rb
```diff
@@ -13,6 +13,6 @@ def create_preservation_event(object, event)
                                               initiating_user:  event['user'],
                                               outcome:          event['outcome'],
                                               software_version: event['software_version'] }]
-    object.save!
+    object.save
```
rotated8: Please open a ticket pointing to this diff so we can fix the affected tests.

spec/actors/hyrax/actors/curate_generic_work_actor_spec.rb
devanshu-m: Removed redlock client stub

app/lib/preservation_events.rb
devanshu-m: Replaced failing system test with unit test to restore the bang save.

Issue #616, opened 2019-09-26
Display all Technical Metadata on View FileSet page


eporter23:
Expand the display of Technical Metadata on the View FileSet page so that it includes all the metadata extracted by FITS for the Preservation Master File

[List of technical metadata fields which may be extracted](https://docs.google.com/spreadsheets/d/1csDcDnJdhB1VLlgJUU95DndXSeIP2j8yPMlErrsVVPw/edit#gid=1836172006)

wulman16:
The work for this ticket is done except for the automated test. I've attempted mocking the characterization service response in the spec with little success. The next step would probably be to try mocking the Solr document itself. There are a couple of examples in the Hyrax spec, such as:
https://github.com/samvera/hyrax/blob/v3.0.0.pre.beta1/spec/views/hyrax/file_sets/show.html.erb_spec.rb
https://github.com/samvera/hyrax/blob/v3.0.0.pre.beta1/spec/presenters/hyrax/file_set_presenter_spec.rb

eporter23:
@wulman16 looks good! I also observed that the extended technical metadata displays on new works created after the deployment, so I'm glad we got this in before migrations are underway.

PR #750, opened 2019-11-12
Display all technical metadata on view fileset page

wulman16:
All required metadata extracted by FITS for the preservation master file now appears on the view fileset page. Characterization metadata already stored in Fedora is now indexed into Solr and shown to the user.

Resolves: #616


app/presenters/curate/file_set_presenter.rb
```diff
@@ -1,5 +1,20 @@
...
+    delegate :well_formed, to: :solr_document
+    delegate :creating_application_name, to: :solr_document
+    delegate :puid, to: :solr_document
+    delegate :date_modified, to: :solr_document
+    delegate :character_set, to: :solr_document
+    delegate :byte_order, to: :solr_document
+    delegate :color_space, to: :solr_document
+    delegate :compression, to: :solr_document
+    delegate :profile_name, to: :solr_document
+    delegate :profile_version, to: :solr_document
```
rotated8: Let's put this into a single delegate block

wulman16:
Before: 
Screen Shot 2019-11-13 at 4 42 55 PM
After:
Screen Shot 2019-11-14 at 10 01 17 AM

PR #760, opened 2019-11-14
Update visibility badges to match Emory expectations

bess:
Update visibility badges to match specifications from Curate user group

Connected to https://github.com/curationexperts/in-house/issues/448


bess:
curate_visibility

PR #763, opened 2019-11-14
Revert "Adds clamby for virus checking"

rotated8:
Reverts emory-libraries/dlp-curate#752


rotated8:
Clamby is falsely detecting every file as a virus, so let's revert this until we can diagnose the issue.

PR #765, opened 2019-11-15
Add additional file types to importer

little9:
This adds the additional file types required
for Yellowbacks to the importer & mapper:

* `service_file`
* `extracted`
* `transcript_file`

This also updates the logic for creating `PreIngestFiles`
so that they will be created for these new types.


app/importers/curate_record_importer.rb
```diff
@@ -21,25 +22,34 @@ def create_upload_files(record)
     uploaded_file_ids = []
     files_to_attach.each do |filename|
```
mark-dce: I know this isn't part of the PR, but I'm finding the naming really hard to sort out here.  It looks like `files_to_attach` are really CSV rows for filesets and a `filename` is really something that contains multiple files - is it really something more like a fileset or event a fieset_row?

PR #766, opened 2019-11-15
Add pcdm_use to mapper and importer

little9:
This allows you to use pcdm_use in the CSV
and maps the values to the terms defined
on `FileSet`.

* 'Primary Content' -> FileSet::PRIMARY
* 'Supplemental Content' -> FileSet::SUPPLEMENTAL
* 'Supplemental Preservation' -> FileSet::PRESERVATION
*  No value - > FileSet::Primary

Connected to https://github.com/curationexperts/in-house/issues/463


app/importers/curate_mapper.rb
```diff
@@ -135,6 +135,30 @@ def administrative_unit
...
+    case normalized_csv_term
+    when ''
+      FileSet::PRIMARY
+    when 'primary content'
+      FileSet::PRIMARY
+    when 'supplemental content'
+      FileSet::SUPPLEMENTAL
+    when 'supplemental preservation'
+      FileSet::PRESERVATION
+    end
```
mark-dce: Isn't [`FileSet::PRESERVATION`](https://github.com/emory-libraries/dlp-curate/blob/8c1e18f29083829aa945d57f95808158f5869ca9/app/models/file_set.rb#L5) just a constant with the string value 'Supplemental Preservation'?  

Would something like this be clearer?
```
VALID_PCDM_USES = ['', FileSet::PRIMARY, FileSet::SUPPLEMENTAL, FileSet::PRESERVATION].map(&:downcase) # this could even go in app/models/file_set.rb

raise "Invalid PCDM use value: #{csv_term}" unless VALID_PCSM_USES.include? csv_term.downcase
csv_term
```

In either case, you could also skip the downcase and just require the user to provide proper capitalization

PR #769, opened 2019-11-18
preferred image derivatives

### LINKED ISSUE #717: Apply preferred image logic to generate thumbnails used by Hyrax at ingest
Look at options to make sure Hyrax/Curate generates its thumbnail(s) for works/FileSets based on our preferred image logic from #411. These thumbs should then be available as needed within Hyrax and Lux displays

View/Edit FileSet thumbnail
View Work - Items thumbnail
Search results thumbnails

alexBLR:
Closes #717


app/actors/hyrax/actors/file_set_actor.rb
```diff
@@ -20,26 +20,28 @@ def initialize(file_set, user)
...
+          file_actor.ingest_file(wrapper!(file: file, relation: relation, preferred: preferred))
           # Copy visibility and permissions from parent (work) to
           # FileSets even if they come in from BrowseEverything
           VisibilityCopyJob.perform_later(file_set.parent)
           InheritPermissionsJob.perform_later(file_set.parent)
         else
-          IngestJob.perform_later(wrapper!(file: file, relation: relation))
+          IngestJob.perform_later(io_wrapper)
         end
+        # file_actor.preferrred_file(preferred, relation)
```
rotated8: Since this line is commented out, can it be removed?

app/jobs/characterize_job.rb
```diff
@@ -25,6 +25,5 @@ def perform(file_set, file_id, filepath = nil)
     file_set.characterization_proxy.save!
     file_set.update_index
     file_set.parent&.in_collections&.each(&:update_index)
-    CreateDerivativesJob.perform_later(file_set, file_id, filepath)
```
rotated8: Want to check my understanding. This is moving to the `file_actor.rb` initializer, because the characterize job only runs on the master file, but we want to run the derivatives job on a file we choose. Correct?

spec/jobs/characterize_job_spec.rb
```diff
@@ -24,7 +24,6 @@
   before do
     allow(FileSet).to receive(:find).with(file_set_id).and_return(file_set)
     allow(Hydra::Works::CharacterizationService).to receive(:run).with(file, filename)
-    allow(CreateDerivativesJob).to receive(:perform_later).with(file_set, file.id, filename)
```
rotated8: This is removed here but not tested somewhere else. Are we missing a place to test this?

spec/controllers/hyrax/downloads_controller_spec.rb
```diff
@@ -144,14 +144,6 @@
...
         end
-
-        context "that isn't persisted" do
-          it "raises an error if the requested file does not exist" do
-            expect do
-              get :show, params: { id: file_set, file: 'thumbnail' }
-            end.to raise_error ActiveFedora::ObjectNotFoundError
-          end
-        end
       end
```
rotated8: Can you say a bit about why this is removed?

app/jobs/characterize_job.rb
alexBLR: Yes. it is correct.

spec/controllers/hyrax/downloads_controller_spec.rb
alexBLR: I am adding it back. Tests seem to pass now.

app/actors/hyrax/actors/file_set_actor.rb
alexBLR: Removed

spec/jobs/characterize_job_spec.rb
alexBLR: added to file_actor

PR #793, opened 2019-11-25
Adds virus-check preservation event

devanshu-m:
* This commit overrides `viruses?` method that comes from
`Hydra::Works::VirusCheck` and is mixed-in through `::Hyrax::FileSetBehavior`.
* It adds the preservation_event to the `viruses?` method and creates a
virus_check event depending on the result returned from the
`file_has_virus?` method in `Hydra::Works::VirusCheckerService`.
* Also, cleans up previous virus_check file since we don't need it with
the hyrax version we are using.


spec/models/file_set_spec.rb
```diff
@@ -85,4 +85,51 @@
       expect(file_set.preservation_event.build).to be_instance_of PreservationEvent
     end
   end
+
+  # This spec is borrowed and modified from `Hydra::Works::VirusCheck`
+  context "with ClamAV" do
+    # We create an empty file_set and walk through virus-checking
+    let :file_set do
+      FileSet.new
+    end
```
rotated8: Please make this a one-liner.

spec/models/file_set_spec.rb
```diff
@@ -85,4 +85,51 @@
...
+      end
+      it 'fails to validate' do
+        expect(file_set.validate).to eq false
+      end
+    end
+
+    context 'with a clean file' do
+      before do
+      end
+
```
rotated8: These lines can be deleted.

PR #803, opened 2019-11-28
Adds Honeybadger

rotated8:
Deploys to arch, test, and prod will report their environment as such, rather than all saying "production"


rotated8:
Requires emory-libraries/cor-cm#5

PR #812, opened 2019-12-04
adds fixity check button on file_set show page

devanshu-m:
* This commit adds a button for running fixity check on demand.
* We modify the create method in the FixityChecksController since
we no longer need to render json, but redirect to the file_set show
page with the result from the fixity check.


devanshu-m:
image

PR #841, opened 2019-12-10
Adds fixity_check preservation_event

devanshu-m:
* Adds fixity_check preservation_event to run_check method in
FixityCheckJob. Takes the ChecksumAuditLog for a file and checks
for the `passed` attribute. If true, creates success event else
failure event.


devanshu-m:
image

PR #870, opened 2019-12-20
Catch failures when edtf-humanize gem is used

wulman16:


spec/services/date_service_spec.rb
```diff
@@ -36,5 +36,9 @@
     it 'provides a string for dates with known day and month but unknown year' do
       expect(service.human_readable_date('XXXX-09-07')).to eq('September 7, year unknown')
     end
+
+    it 'provides a string for dates with unexpected YMD formatting' do
+      expect(service.human_readable_date('2019-XX-07')).to eq('2019-XX-07')
```
devanshu-m: Is this the test to check if it returns raw date if humanize can't be called on the date?

wulman16: Yes

PR #872, opened 2019-12-23
Save to Solr so fields are available to Lux for facets

maxkadel:
Still unsure about date facets


maxkadel:
Closing for now, still working on dates.

PR #881, opened 2020-01-06
Hard code sidekiq workers

SolomonHD:
Adjustments to the sidekiq config to avoid using the SIDEKIQ_WORKERS variable which was not being recognized.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,14 +1,14 @@
 ---
-  :queues:
-    - [ingest, 4]
-    - [batch, 2]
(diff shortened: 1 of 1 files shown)


SolomonHD:
The env var SIDEKIQ_WORKERS is not being recognized by the ruby app. I've tried experimenting with dotenv and loading dotenv at different times in the application boot sequence and it still didn't work. The app has always been using the old default of 5 workers.

This PR rewrites the sidekiq.yml file so that we don't use SIDEKIQ_WORKERS.

PR #895, opened 2020-01-13
Fix honeybadger config for development environment

maxkadel:
When running in development environment, got following message, added safe operators to allow to run in development environment.

`/Users/max/projects/dlp-curate/config/honeybadger.yml:3:in '<main>': undefined method 'partition' for nil:NilClass (Honeybadger::Config::ConfigError)`


rotated8:
@maxkadel Is this still a problem with master?

rotated8:
Closing per Lux Standup.

PR #914, opened 2020-01-26
Adds manifest caching

devanshu-m:
* Manifest creation for large book objects times out. This PR implements caching of manifests. If there is a cached manifest for an object, UV will use that, else it'll create a new manifest and cache it for future use.
* There is also a rake task that creates manifests for all the works (or a single work) and caches them.
Rake task: `RAILS_ENV=*env* bundle exec rails curate:works:create_manifest` OR `RAILS_ENV=*env* bundle exec rails curate:works:create_manifest work=*work_id*`

### DIFF
--- a/app/presenters/hyrax/curate_generic_work_presenter.rb
+++ b/app/presenters/hyrax/curate_generic_work_presenter.rb
@@ -7,5 +7,25 @@ module Hyrax
     CurateGenericWorkAttributes.instance.attributes.each do |key|
       delegate key.to_sym, to: :solr_document
     end
+
+    # [Hyrax-overwrite] We might not always have a request and a `base_url`,
+    # therfore, we are using our CurateManifestHelper and passing in a hardcoded
+    # host for creation of manifest_url
+
+    def manifest_helper
+      @manifest_helper ||= if request.nil?
+                             CurateManifestHelper.new
+                           else
+                             ManifestHelper.new(request.base_url)
+                           end
+    end
+
+    def manifest_url
+      if request.nil?
+        manifest_helper.polymorphic_url([:manifest, self], host: "http://#{ENV['HOSTNAME'] || 'localhost:3000'}")
+      else
+        manifest_helper.polymorphic_url([:manifest, self])
+      end
+    end
   end
 end
--- a/spec/lib/create_manifest_spec.rb
+++ b/spec/lib/create_manifest_spec.rb
@@ -21,9 +21,11 @@ RSpec.describe CreateManifest, :clean do
   before do
     allow(CurateGenericWork).to receive(:all).and_return([generic_work])
     allow(SolrDocument).to receive(:find).and_return(solr_document)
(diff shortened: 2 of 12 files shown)


spec/lib/create_manifest_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+  let(:presenter) { described_class.new(solr_document, ability, request) }
+
+  before do
+    allow(CurateGenericWork).to receive(:all).and_return([generic_work])
+    allow(SolrDocument).to receive(:find).and_return(solr_document)
+  end
+
+  it 'creates manifest and saves to tmp' do
+    described_class.process
```
rotated8: Can we assert it is not present before we create it? It may be cautious to delete it if it exists, then assert it isn't present.

spec/presenters/hyrax/curate_generic_work_presenter_spec.rb
```diff
@@ -5,7 +5,27 @@
...
+  let(:presenter) { described_class.new(solr_document, ability, request) }
+
+  describe '#manifest_url' do
+    subject { presenter.manifest_url }
+    before do
+      allow(request).to receive(:host).and_return 'example.org'
+      allow(request).to receive(:base_url).and_return 'http://example.org'
+    end
+
+    it { is_expected.to eq 'http://example.org/concern/curate_generic_works/888888/manifest' }
```
rotated8: Can we also test this with host/base_url unset? I'd like to see the localhost piece working.

app/presenters/hyrax/curate_generic_work_presenter.rb
```diff
@@ -7,5 +7,25 @@ class CurateGenericWorkPresenter < Hyrax::WorkShowPresenter
...
+      @manifest_helper ||= if request.nil?
+                             CurateManifestHelper.new
+                           else
+                             ManifestHelper.new(request.base_url)
+                           end
+    end
+
+    def manifest_url
+      if request.nil?
+        manifest_helper.polymorphic_url([:manifest, self], host: "http://#{ENV['HOSTNAME'] || 'localhost:3000'}")
```
rotated8: What happens if a developer isn't running their server on port 3000?

spec/lib/create_manifest_spec.rb
devanshu-m: I have updated the test.

spec/presenters/hyrax/curate_generic_work_presenter_spec.rb
devanshu-m: I have updated the test with new contexts.

app/presenters/hyrax/curate_generic_work_presenter.rb
devanshu-m: I have updated the readme to reflect this.

rotated8:
@bess Here's the caching PR

PR #934, opened 2020-02-01
IIIF endpoint / proxy

bess:
* Create a iiif endpoint that will proxy requests to the IIIF endpoint described by PROXIED_IIIF_SERVER_URL
* This lays the groundwork necessary for rewriting IIIF requests depending on the user's status and the object's visibility settings

Connected to https://github.com/emory-libraries/dlp-lux/issues/260


spec/controllers/iiif_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  describe "a request for a public object" do
+    let(:expected_iiif_url) { 'http://127.0.0.1:8182/iiif/2/abc123/full/full/0/default.jpg' }
+
+    before do
+      ENV['PROXIED_IIIF_SERVER_URL'] = 'http://127.0.0.1:8182/iiif/2'
+    end
+
+    it "does not alter the iiif request" do
+      get :show, params: params
```
rotated8: This `get` fails in CI, because the connection to 127.0.0.1:8182 is refused. Should we mock a response?

maxkadel: What confuses me is that the previous CI runs passed, until the "remove cruft" edit, which seems to only remove commented out lines, so I'm confused.

bess:
@rotated8 Fixed! Take another look when you get a chance, please?

spec/controllers/iiif_controller_spec.rb
rotated8: Clicking on the checks by the commits, it appears CI did not run for them. Currently, CI only runs against PRs, in an effort to reduce the CI queue when we're all pushing work.

I agree that it is confusing that GitHub appears to show that it not only ran, but passed.

PR #945, opened 2020-02-04
Adds preferred_file method for file_set instance

devanshu-m:
* This commit adds a new file_set instance method `preferred_file`
which returns the preferred_file for the file_set.
* Instances where `preferred_file` logic was being used is replaced
with a call to this new instance method.
* Tests are updated as well.

### DIFF
--- a/app/models/file_set.rb
+++ b/app/models/file_set.rb
@@ -74,6 +74,17 @@ class FileSet < ActiveFedora::Base
     result
   end
 
+  def preferred_file
+    preferred = if service_file.present?
+                  :service_file
+                elsif intermediate_file.present?
+                  :intermediate_file
+                else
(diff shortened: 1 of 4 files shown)


app/models/file_set.rb
```diff
@@ -74,6 +74,17 @@ def viruses?
...
 
+  def preferred_file
+    preferred = if service_file.present?
+                  :service_file
+                elsif intermediate_file.present?
+                  :intermediate_file
+                else
+                  :preservation_master_file
+                end
+    preferred
```
rotated8: Is the `preferred = ` ... `preferred` portion required? Will Ruby do an implicit return without it?

devanshu-m: I have updated the method. Please take a look.

PR #949, opened 2020-02-04
Implements FileSetCleanUp as a background job

devanshu-m:
* This commit changes how the rake task runs the file_set cleanup
stuff by moving it into a background job. This way we can run the
rake task which will exit almost immediately after kicking off the
FileSetCleanUpJob that runs as a background job in sidekiq.


rotated8:
This will be approved when #945 gets merged, and this PR's target branch is updated.

devanshu-m:
@rotated8 This branch has been rebased with master and ready to be reviewed.

PR #967, opened 2020-02-10
Adds a job to check if binaries exist in s3

devanshu-m:
* This commit adds a job that will take a bucket as input and check that
files in file_sets exist in s3 or not.
* There is a rake task that kicks off this job: `RAILS_ENV=*env* bundle
exec rails curate:file_sets:check_binaries bucket=*s3-bucket-name*
file_set=*optional*`

### DIFF
--- a/app/jobs/check_binaries_job.rb
+++ b/app/jobs/check_binaries_job.rb
@@ -11,9 +11,7 @@ class CheckBinariesJob < Hyrax::ApplicationJob
         fs = FileSet.find(file_set_id)
         check_binary(fs, csv)
       else
-        FileSet.all.each do |file_set|
-          check_binary(file_set, csv) unless file_set.files.empty?
-        end
+        check_all_file_sets
       end
     end
   end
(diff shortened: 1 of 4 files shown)


app/jobs/check_binaries_job.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private
+
+    def check_binary(file_set, csv)
+      file_set.files.each do |file|
+        next if file.digest.empty?
+        sha1 = file.digest.first.to_s.partition("urn:sha1:").last
+        next if @bucket.object(sha1).exists?
+        csv << [file&.id]
```
rotated8: It'd be nice to get some more information here. Parent Work ID, Fileset's ID, and SHA1?

devanshu-m: I have updated the csv results. Please take a look.

app/jobs/check_binaries_job.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+    end
+  end
+
+  private
+
+    def check_binary(file_set, csv)
+      file_set.files.each do |file|
+        next if file.digest.empty? || check_existence_in_s3(file)
+        csv << [file_set.member_of_work_ids.join(','), file_set.id, file&.id, @sha1]
```
rotated8: We shouldn't join this with a comma since we're storing this in a CSV. We should be able to use a '|', though.

PR #973, opened 2020-02-11
IIIF controller must proxy info.json requests

bess:
It isn't enough that the IiifController proxies image requests to
cantaloupe -- it needs to also proxy info.json requests. This PR adds
that feature.

BUGFIX: Only add an extra slash when it is needed to construct a IIIF url

Connected to https://github.com/emory-libraries/dlp-lux/issues/260


bess:
Screen shot of this working on curate-test:

proxy_logs

PR #976, opened 2020-02-12
displays preferred image in work show page iiif

### LINKED ISSUE #681: Display Preferred Image on View Work Page
Further splits out the display of preferred image logic for the Universal Viewer and thumbnails on the View Work page.

- [x] Change manifest generation to use the preferred image from #942

See [mockup for View Work display](https://docs.google.com/drawings/d/1Hl01D-ITNFkDhdnlCNRRZrGCTYvrdzFCcASZ7Etg1rY/edit?usp=sharing)

See also:
Original ticket from DCE co-sprint: #411
FileSet work developed in #442

alexBLR:
Closes #681 

After multiple attempts writing system tests for iframe for iiif there is no way to identify which file it is displaying within the UI. Iframe point directly to the work manifest. The image is drawn with Canvas.

Hyrax only has tests for display_image in a file_set_presenter_spec.rb. This would be for a different ticket where we need to display preferred image on a fille_set page with UV

### DIFF
--- a/app/presenters/hyrax/displays_image.rb
+++ b/app/presenters/hyrax/displays_image.rb
@@ -15,11 +15,12 @@ module Hyrax
     def display_image
       return nil unless ::FileSet.exists?(id) && solr_document.image? && current_ability.can?(:read, id)
       # @todo this is slow, find a better way (perhaps index iiif url):
-      original_file = ::FileSet.find(id).original_file
+      file_set = ::FileSet.find(id)
+      preferred_file = file_set.send(file_set.preferred_file)
       @request_base_url = request_base_url
 
       url = Hyrax.config.iiif_image_url_builder.call(
-        original_file.id,
+        preferred_file.id,
         @request_base_url,
         Hyrax.config.iiif_image_size_default
       )
@@ -27,7 +28,7 @@ module Hyrax
       IIIFManifest::DisplayImage.new(url,
                                      width:         640,
                                      height:        480,
-                                     iiif_endpoint: iiif_endpoint(original_file.id))
+                                     iiif_endpoint: iiif_endpoint(preferred_file.id))
     end
 
     private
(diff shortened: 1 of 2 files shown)


spec/presenters/hyrax/file_set_presenter_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+require 'rails_helper'
+require 'iiif_manifest'
+RSpec.describe Hyrax::FileSetPresenter do
+  let(:ability) { instance_double(Ability) }
+  let(:user) { instance_double(user_key: 'sarah') }
```
devanshu-m: Do we need these lines? Are we using them in the test?

spec/presenters/hyrax/file_set_presenter_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  let(:file_set) { FactoryBot.create(:file_set) }
+  let(:pmf)      { File.open(fixture_path + '/book_page/0003_preservation_master.tif') }
+  let(:sf)       { File.open(fixture_path + '/book_page/0003_service.jpg') }
+  let(:imf)      { File.open(fixture_path + '/book_page/0003_intermediate.jp2') }
+  let(:read_permission) { true }
```
devanshu-m: Are we using this?

app/presenters/hyrax/displays_image.rb
```diff
@@ -15,7 +15,8 @@ module DisplaysImage
     def display_image
       return nil unless ::FileSet.exists?(id) && solr_document.image? && current_ability.can?(:read, id)
       # @todo this is slow, find a better way (perhaps index iiif url):
-      original_file = ::FileSet.find(id).original_file
+      file_set = ::FileSet.find(id)
+      original_file = file_set.send(file_set.preferred_file)
```
devanshu-m: Since this is now preferred_file and not original_file. Can we rename this to something more relatable?

PR #982, opened 2020-02-13
Sets the CORS header for the IIIF proxy

rotated8:
When the Access-Control-Allow-Origin header is missing, the content sent is not loaded by UV.


rotated8:
@bess I should note, I have not tested this against Lux.

PR #1004, opened 2020-02-17
Redirects to origin when signing in via shib

devanshu-m:
* In this commit, we redirect the user to omniauth request's origin
if it exists, else will redirect to the dashboard. Refer:
https://github.com/omniauth/omniauth/wiki/Saving-User-Location

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -5,7 +5,10 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
     @user = User.from_omniauth(request.env["omniauth.auth"])
 
     if @user.persisted?
-      sign_in_and_redirect @user
+      sign_in @user
+      # This will help the user go back to where they came.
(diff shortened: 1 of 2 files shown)


bess:
This looks great, and is probably a good thing to implement in general, but does not seem to solve the case of the missing manfests:

omniauth_deployed

manifest_still_not_loading

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -5,7 +5,10 @@ def shibboleth
     @user = User.from_omniauth(request.env["omniauth.auth"])
 
     if @user.persisted?
-      sign_in_and_redirect @user
+      sign_in @user
+      # This will help the user go back to where they came.
+      # Refer: https://github.com/omniauth/omniauth/wiki/Saving-User-Location
+      redirect_to request.env["omniauth.origin"] || '/dashboard'
```
rotated8: Rather than using `'/dashboard'`, we should use a named route.

PR #1011, opened 2020-02-19
Adds a way to check if assign_id for works exists

devanshu-m:
* This commit modifies the assign_id method for the work, file_set,
and collection.
* Sometimes the minter goes into a locked state when it tries
creating resources using IDs that are already taken or gone in
Fedora. In order to make sure that a resource is only created on an
unused valid ID, we loop through IDs until we find one that is not
previously assigned to any resource in Fedora.


app/lib/identifier.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Identifier
```
rotated8: Can we call this `CurateIdentifier`, to be clearer?

PR #1052, opened 2020-02-28
Updates bundler to 2.1.2

fnibbit:
This fixed a problem for me locally, is it ok to upgrade bundler?


rotated8:
@SolomonHD How does Ansible install bundler?

SolomonHD:
Ruby is installed by this role: https://github.com/emory-libraries/emorylib_role_install_ruby_scl/blob/master/tasks/main.yml

It installs whatever latest bundler is available. I believe that's 2.1.4 currently.

PR #1062, opened 2020-03-02
Adjust region requests for low_res visibility

bess:
When a user requests a region for a low_res object, ensure the region is
limited to the max pixels permitted.

Connected to https://github.com/emory-libraries/dlp-lux/issues/313


bess:
Not quite right yet.

PR #1064, opened 2020-03-03
 Limit deep zoom by limiting the region values

bess:
For low_res visibility images, do not allow region values that would
enable deepest levels of zooming.

Default value is 800 pixels. This can be adjusted via an environment
variable MIN_TILE_SIZE_FOR_LOW_RES.

Connected to https://github.com/emory-libraries/dlp-lux/issues/313

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,2 +1,0 @@
-# Use local database auth instead of Shibboleth
-export DATABASE_AUTH=true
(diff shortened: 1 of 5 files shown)


bess:
Before, direct iiif request for deep zoom:

before_zoom_limited

After, same request:

limited_zoom

PR #1105, opened 2020-03-23
Add new env var and capistrano task to fix banner persistence

SolomonHD:
This branch will add a new optional environment variable called BRANDING_PATH, this will control where curate save banners and other branding images after the ingest process. Additionally, an optional capistrano variable called branding_symlink_path will create a symlink at public/branding to where the path is defined.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ flycheck_*.el
 /vendor/bundle
 
 .env.development
+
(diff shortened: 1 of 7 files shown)


SolomonHD:
Please reach out to me before this is about to be deployed, I'll have to make some changes on the server to ensure the banners are in the right spot.

PR #1109, opened 2020-03-25
Indexes banner path to the Solr query.

bwatson78:
1. curate_collection_indexer.rb: adds the custom field to the indexer with two methods that pluck
the needed path from the document object.
2. collection_indexer_spec.rb: creates the tests for the new index field.

### DIFF
--- a/app/indexers/curate_collection_indexer.rb
+++ b/app/indexers/curate_collection_indexer.rb
@@ -22,7 +22,7 @@ class CurateCollectionIndexer < Hyrax::CollectionIndexer
 
   def banner_path
     cbi = branding_details
-    cbi.nil? || cbi&.local_path&.nil? ? '' : path_sanitized(cbi.local_path)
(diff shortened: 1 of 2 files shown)


app/indexers/curate_collection_indexer.rb
```diff
@@ -11,11 +11,25 @@ def generate_solr_document
...
+    cbi = branding_details
+    cbi.nil? || cbi&.local_path&.nil? ? '' : path_sanitized(cbi.local_path)
+  end
+
+  def branding_details
+    CollectionBrandingInfo.find_by_collection_id_and_role object.id, 'banner'
+  end
+
+  def path_sanitized(path)
+    path.gsub ::Rails.root.to_s + '/public', ''
```
rotated8: The path is [not guaranteed](https://github.com/emory-libraries/dlp-curate/blob/master/config/initializers/hyrax.rb#L212) to start with Rails.root and public.

PR #1134, opened 2020-04-03
DO NOT MERGE: Thumbs for low res revamp

maxkadel:


maxkadel:
Taking a different approach for thumbnails, via the IIIF controller

PR #1136, opened 2020-04-06
Moves process of creating manifest to a job

devanshu-m:
* We are now starting a background job through a rake task when we create manifests for works. There are also changes to the way preferred_file's id is fetched to include better error handling when file_set's pmf is missing for some reason. This will make sure the background job doesn't error out and keeps running.


devanshu-m:
@rotated8 I have done a rebase with master. Can you review please?

PR #1137, opened 2020-04-07
Disable SolrSuggest

devanshu-m:
* SolrSuggest may be responsible for slowdown in ingest pace. Disabling SolrSuggest might help to decrease time taken to create and add a work to a collection. Refer:
https://docs.google.com/presentation/d/1Jj0aC8LhfbcWxe0g8bBhTj1wxKMbkU6v11iJRBKNV3E/edit?copiedFromTrash#slide=id.g44d37b2273_2_0
and
https://docs.google.com/presentation/d/1Jj0aC8LhfbcWxe0g8bBhTj1wxKMbkU6v11iJRBKNV3E/edit?copiedFromTrash#slide=id.g44d37b2273_2_7
and
https://github.com/samvera/active_fedora/pull/1311/files

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -314,6 +314,9 @@
       </analyzer>
     </fieldType>
 
+<!-- SolrSuggest disabled due to performance penalties -->
+<!--
+
     <fieldType class="solr.TextField" name="textSuggest" positionIncrementGap="100">
        <analyzer>
           <tokenizer class="solr.KeywordTokenizerFactory"/>
@@ -322,6 +325,7 @@
           <filter class="solr.RemoveDuplicatesTokenFilterFactory"/>
        </analyzer>
     </fieldType>
+-->
 
     <!-- charFilter + WhitespaceTokenizer  -->
     <!--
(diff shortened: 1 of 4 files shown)


bess:
@devanshu-m I 100% endorse this change in solr. However, could you check with @SolomonHD whether making the change to this solr config will actually affect the solr config we're running on the servers? Maybe it will, but I would not assume that without discussing it.

SolomonHD:
This will not affect the solr-cor servers, I believe the solr in this repo only controls the local solr when you run this app in development mode.

devanshu-m:
What do you suggest then? @SolomonHD How do we get these changes on the solr-cor servers?

mprefer:
Changes need to be made to schema.xml and solrconfig.xml that reside in Zookeeper.
They need to be copied from ZK to local disk, modified, then copied back into ZK.

SolomonHD:
Yes this is probably best done manually at this point, the scripts haven't been ran on these boxes in almost a full year. Originally the schema/solrconfig were templated from [here](https://github.com/curationexperts/uclalib_role_solr7/blob/add-solr-cloud/files/hyrax_solr-conf/schema.xml) and [here](https://github.com/curationexperts/uclalib_role_solr7/blob/add-solr-cloud/templates/hyrax_solrconfig/solrconfig_xml.j2) respectively.

SolomonHD:
@devanshu-m Since the /conf folder isn't being used what do you think about deleting it for simplicity?

PR #1161, opened 2020-04-17
Increase cookie security by ensuring that the value is not for an old, shared cookie

maxkadel:


maxkadel:
using a different branch

PR #1173, opened 2020-04-28
Info method bug

maxkadel:


app/controllers/iiif_controller.rb
```diff
@@ -139,9 +139,28 @@ def info
...
+      @info_public_iiif = rewrite_iiif_base_uri(@info_original)
+      send_data @info_public_iiif, type: 'application/json', x_sendfile: true, disposition: 'inline'
+    elsif response.status == 404
+      error_message = "Request to #{iiif_url} resulted in 404 Not Found response, likely a file missing in Fedora"
+      # send error to Honeybadger
+      Honeybadger.notify(error_message)
+      render_404
+    else
+      error_message = "Request to #{iiif_url} resulted in #{response.status}"
+      Honeybadger.notify(error_message)
```
bess, line 158: I think we want to render the 404 page here too, don't we?

maxkadel:
Currently rendering the html as text, rather than as html 
image

PR #1201, opened 2020-05-28
Adds preservation workflow info to status block

devanshu-m:
* We are adding preservation workflow information for a work to the status block. The information is populated if it is supplied.

Output examples:
image
image

### DIFF
--- a/app/views/hyrax/base/_preservation_status.html.erb
+++ b/app/views/hyrax/base/_preservation_status.html.erb
@@ -12,24 +12,20 @@
         <b><%= t('hyrax.base.show.depositor') %>: </b><%= presenter.depositor.first %>
       </div>
       <div>
-        <% unless presenter.preservation_workflows.empty? %>
-          <% presenter.preservation_workflows.each do |pwf| %>
-            <b><%= pwf["workflow_type"] %>:</b>
-              <% if pwf.values.compact.count > 1 %>
-                <% pwf.each do |k, v| %> 
-                  <div><%= "#{k.split('_').map(&:capitalize).join(' ').remove('Workflow ')}: #{v}" if v && k != "workflow_type" %></div>
-                <% end %>
-              <% else %>
-                <div>No information supplied</div>
-              <% end %>
+        <!-- iterate through workflows -->
+        <% presenter.preservation_workflows.each do |pwf| %>
+        <!-- display workflow type-->
+        <b><%= pwf["Type"] %>:</b>
+          <!-- display key value pairs only if workflow details exist -->
+          <% unless pwf.values.count > 1 %>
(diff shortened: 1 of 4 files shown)


app/views/hyrax/base/_preservation_status.html.erb
```diff
@@ -1,7 +1,7 @@
 <!-- Preservation status block -->
 <div class="container-fluid">
   <div class="row">
-    <div class="col-xs-4 col-sm-4 col-md-4 col-lg-4">
+    <div class="col-xs-6 col-sm-6 col-md-6 col-lg-6">
```
wulman16, line 4: Something I didn't catch last time: since the column proportion is the same for all breakpoints, you can accomplish the same thing with `class="col-6"`

app/views/hyrax/base/_preservation_status.html.erb
```diff
@@ -11,6 +11,25 @@
       <div>
         <b><%= t('hyrax.base.show.depositor') %>: </b><%= presenter.depositor.first %>
       </div>
+      <div>
+        <% unless presenter.preservation_workflows.empty? %>
+          <% presenter.preservation_workflows.each do |pwf| %>
+            <b><%= pwf["workflow_type"] %>:</b>
+              <% if pwf.values.compact.count > 1 %>
+                <% pwf.each do |k, v| %> 
+                  <div><%= "#{k.split('_').map(&:capitalize).join(' ').remove('Workflow ')}: #{v}" if v && k != "workflow_type" %></div>
```
wulman16, line 20: Can we move more of the logic out of the view and into the presenter?

app/views/hyrax/base/_preservation_status.html.erb
devanshu-m: I tried using that last time and it didn't work for me. Are there any other alternatives I can try?

app/views/hyrax/base/_preservation_status.html.erb
devanshu-m: Ok, I will try to move around some logic.

app/views/hyrax/base/_preservation_status.html.erb
wulman16: Oh right, Hyrax uses Bootstrap 3. You should be able to use `class="col-xs-6 col-s-6"` in that case. https://getbootstrap.com/docs/3.4/css/#grid-options

app/views/hyrax/base/_preservation_status.html.erb
devanshu-m: @wulman16 I have moved most of the logic into the presenter.

Issue #1204, opened 2020-06-01
Revision for Curate ticket 1026


eporter23:
Revision for work completed in Curate #1026 : please add a break or margin between the different Workflow entries so they are easier to read, as shown in the mockup:

![Screen Shot 2020-06-01 at 2.28.45 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/89e88b97-29ee-45f1-8e35-07161f707912)

Please also remove the duplicated workflow metadata from the standard metadata display (highlighted in red below) now that it displays in the Preservation Status block:

![Screen Shot 2020-06-01 at 2.37.17 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/faab3b7f-7ea4-4f0b-a176-707e14603452)

bwatson78:
![Screen Shot 2020-06-02 at 9.14.11 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/7e0f2266-5db0-4bdc-9d7b-4a2d72cb2b9b)

PR made: https://github.com/emory-libraries/dlp-curate/pull/1208

PR #1208, opened 2020-06-02
Removes workflow metadata from main object block and spaces out Acces…

bwatson78:
…sion and Ingest details in Preservation Block #1204.
- curate_generic_works_controller.rb: extracts unneeded instance variable definition.
- _preservation_status.html.erb: adds line breaks and forgotten div ending.
- app/views/hyrax/base/show.html.erb: strips outputting logic no longer required.
- spec/system/viewing_a_work_spec.rb: takes out tests for elements that no longer exist.


bwatson78:
Screen Shot 2020-06-02 at 9 14 11 AM

PR #1209, opened 2020-06-02
Display preservation master file in files table

wulman16:
- If a preservation master file for a fileset exists, display it in the files table on the fileset show page
- Move the files table to the desired place on the fileset show page


wulman16:
Screen Shot 2020-06-02 at 10 41 26 AM

Issue #1036, opened 2020-02-24
Check User Activity table in revised View FileSet page layout


eporter23:
See [wireframe - section 8](https://docs.google.com/drawings/d/1A--1khkQlaDndQVu22cOKQjZTWRyC-tfFBbzaqlrM7k/edit)
- [x] Confirm that the Hyrax-generated table still works
- [x] Confirm that the table displays in the expected order of page elements

bwatson78:
![Screen Shot 2020-06-03 at 11.00.47 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/a54f0e56-c16a-4ae8-a821-378600b8337d)

PR made: https://github.com/emory-libraries/dlp-curate/pull/1214

bwatson78:
@eporter23 I changed the name as you mentioned, and the table reflected that change (after refresh).

![Screen Shot 2020-06-03 at 11.11.19 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/405d9c5d-399b-4ee0-8455-507c2ca6a317)

eporter23:
@bwatson78 looks good!

eporter23:
@bwatson78 looks good in arch and test. FYI, I also wrote up a new, unrelated ticket for this User Activity table because while testing I noticed that the activity data didn't seem to get restored as part of our backup and restoration process.

eporter23:
@bwatson78 this is good to go for prod when we next deploy

PR #1214, opened 2020-06-03
Repositions and checks function of Activity table #1036.

bwatson78:
- app/views/hyrax/file_sets/show.html.erb: moves table and adds header.
- config/locales/hyrax.en.yml: localizes header text.
- spec/system/show_file_spec.rb: inserts testing for new header and table.


bwatson78:
Screen Shot 2020-06-03 at 11 00 47 AM

PR #1222, opened 2020-06-04
Add fileset ID to fileset show page

wulman16:
- Adds fileset ID to fileset show page
- Adjusts display of fileset category and parent work on fileset show page


wulman16:
Screen Shot 2020-06-04 at 10 27 26 AM

Issue #1037, opened 2020-02-24
Generate new Preservation Events listing on View FileSet page


eporter23:
See mockup - [section 7](https://docs.google.com/drawings/d/1A--1khkQlaDndQVu22cOKQjZTWRyC-tfFBbzaqlrM7k/edit)

- [ ] Create new section and heading to the page output
- [ ] Loop and pull data from Fedora; populate table

To do: revise mockup with alternate layouts as needed to get table contents to fit

Discussion: pull all of this from Fedora for now; if performance is a problem, look into indexing it

devanshu-m:
@bwatson78 moved it by mistake

bwatson78:
@eporter23 Unfortunately, I cannot successfully save files on my local environment, so I cannot generate a screenshot of this working. Rspec tests are passing though, so I expect it to work in test/arch.

bwatson78:
@eporter23 A reindex is needed for all filesets after this is deployed.

eporter23:
@bwatson78 I created a couple of new FileSets in arch and test so I could test your work before the reindex. Looks great! Can you please ensure that the table rows break properly so that there is one event per row?

[arch object](https://curate-arch.library.emory.edu/concern/parent/6214mw6mc9-cor/file_sets/003cz8w9j9-cor)
[test object](https://curate-test.library.emory.edu/concern/parent/4300zpc8g1-cor/file_sets/262s4mw6x0-cor)

![Screen Shot 2020-06-05 at 12.35.32 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/63b27811-bcfa-4d64-beb1-7f2fae8f970e)

bwatson78:
@eporter23 Oops, will fix that today.

bwatson78:
PR made for fix: https://github.com/emory-libraries/dlp-curate/pull/1237

![Screen Shot 2020-06-05 at 1.12.11 PM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/b40f153b-d739-4611-b089-87d665e3e8ea)

bwatson78:
![Screen Shot 2020-06-05 at 1.12.21 PM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/aa41aca8-9f65-4dd5-83d9-50319c4b8c69)

eporter23:
[...]

PR #1224, opened 2020-06-04
Creates new Preservation Events table in FileSet view #1037.

bwatson78:
Also refactors preservation events work on Generic Works Page.
- file_set_indexer.rb: adds preservation_event into solr_doc.
- curate_generic_work_indexer.rb: refactors method that provides the failed events json.
- preservation_event.rb: creates two new json mapping methods.
- file_set_presenter.rb: delegates :preservation_event to the solr_document.
- _preservation_events.html.erb: generates partial for the preservation events table.
- show.html.erb: calls the new partial from the show page.
- config/locales/hyrax.en.yml: institutes multiple headers for use in the new partial.
- spec/*: creates and refactors tests for the new partial.


wulman16:
@bwatson78 Do you have a screenshot of the table?

bwatson78:
@wulman16 My local environment will not save an actual file, so it produces no PreservationEvent. The only thing I can show is the header and a message that there are no events.

PR #1277, opened 2020-06-22
Corrects fact that fileset with alternative is is never saved.

bwatson78:
- spec/lib/iiif_url_builder_service_spec.rb: found that the alternate id fileset was not being saved, so the pulling of the fileset object in dlp-curate/app/lib/iiif_url_builder_service.rb was returning an error. This saves the fileset and passes all errors.

### DIFF
--- a/spec/lib/iiif_url_builder_service_spec.rb
+++ b/spec/lib/iiif_url_builder_service_spec.rb
@@ -4,9 +4,6 @@ require 'rails_helper'
 
 RSpec.describe IiifUrlBuilderService, :clean do
   let(:iiif_builder_service) { described_class.new(file_set_id: file_set.id, size: '260,') }
-  let(:file_set_alt) do
-    FactoryBot.create(:file_set, id: file_set_id_base, user: user, title: ['Some title'])
-  end
   let(:iiif_builder_service_with_alternate_id) { described_class.new(file_set_id: file_set_id, size: '260,') }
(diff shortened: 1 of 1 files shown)


devanshu-m:
What is an alternate id fileset and why do we need it? I apologize, I don't understand. Can you please point me to the code that is making us add this to the test/reason for tests failing if this isn't present? I tried looking in `app/lib/iiif_url_builder_service.rb` but couldn't find any references. Can we discuss this as a group on zoom?

bwatson78:
Sure. This code is for `dlp-curate/spec/lib/iiif_url_builder_service_spec.rb`, line 63. It was throwing an error because, on line 8 in `dlp-curate/app/lib/iiif_url_builder_service.rb`, the finder couldn't return an object with that id. I found that the only fileset that was saved was this one:
```
let(:file_set) do
    FactoryBot.create(:file_set, user: user, title: ['Some title'])
  end
```
That will always have a dynamically created id, so the code in line 63 will never work correctly (it's trying to find an object with the specific id.) So I made an alternate fileset with an implicitly named `id`. The code on line 20 of the spec ensures that the fileset saves by attaching a file. It was not reflecting in the `FileSet.all` call until I inserted that line.

devanshu-m:
Ok, in that case, why can't we just assign `file_set_id_base` as the id to the already existing file_set on L#15?

bwatson78:
Will do.

PR #1280, opened 2020-06-23
Corrects errors caused by preservation workflow field. 

bwatson78:
- app/models/solr_document.rb and app/models/solr_document.rb: swaps field ending with _sim instead of _tesim.


bwatson78:
After reviewing, we've determined that Solr has interpreted our `preservation_workflow_terms_sim` as `preservation_workflow_terms_tesim` when it comes to indexing those fields, so `tesim` call is the right change.

PR #1297, opened 2020-06-26
Adds new Preservation Event Details page.

bwatson78:
- app/controllers/event_details_controller.rb: creates action to display new page.
- app/models/event_details_ability.rb: inserts an ability model that adopts the same terms as Manifest page.
- app/views/event_details/event_details.html.erb: initiates the new erb showing the header.
- app/views/hyrax/base/_preservation_status.html.erb: injects the link for More Event Details.
- config/locales/*: institutes translation for headers and links.
- config/routes.rb: creates route for new page.
- spec/support/shared_examples_for_viewing_a_work.rb: DRYs up the constant use of the multiple has_content method.
- spec/system/viewing_a_work_spec.rb: refactors previous tests and adds one to check for out new link.
- spec/system/viewing_a_works_event_details_spec.rb: checks that page is loading and has a header.


bwatson78:
Screen Shot 2020-06-26 at 2 25 29 PM
Screen Shot 2020-06-26 at 2 24 19 PM

PR #1313, opened 2020-07-02
Include additional fields in dashboard search

wulman16:
- The metadata fields included in the dashboard search bar are now brought in line with the fields searched in the main search bar. The following desired fields are explicitly tested:
    - `deduplication_key`
    - `other_identifiers`
    - `id`
    - `system_of_record_ID`
    - `emory_ark`
- The values for three metadata fields in the `work_with_full_metadata` factory are made longer and unique so that their search functionality can be tested more accurately.


rotated8:
@wulman16 Can you reopen this against the main branch?

PR #1315, opened 2020-07-02
Fixes RDF::JSON::LD issue

devanshu-m:
* When calling `as_json` or `to_json` on an object, sometimes `RDF::JSON` already exists, which leads to name resolution failure. To stop this from happening, we are patching `JSON::LD` module with `::` in the beginning to ensure functions are performed at top level without `RDF` module.
Refer: https://github.com/samvera/valkyrie/pull/810


rotated8:
@devanshu-m I know this is a draft, but can you open this against the main branch, please?

Issue #1265, opened 2020-06-19
Enable remaining generic work fields for CSV importer


eporter23:
When we worked with DCE to implement the CSV bulk import tool, they prioritized enabling fields for which we had representative data populated for testing.

As we move into preparing for new collections, there are still 29 fields in our Curate Generic Work model that are not yet supported for bulk import via CSV that need to be mapped for import.

[List of fields not yet implemented](https://docs.google.com/spreadsheets/d/1Z0ebRLaPa4bEJayvTiqwU4ulFglWrpQPgVX6r6v29uY/edit#gid=714216589&fvid=282953260)

To confirm: if some fields require a URI input in the CSV (this is the case for any locally defined vocabularies for which the form displays a text value, but the URI value is stored in Fedora and SOLR).

Already implemented examples:

- rights_status
- content_type

Not yet implemented, which may follow this pattern:
- re_use_license

Other URI fields do not utilize a locally configured vocabulary:
- related_publications
- related_datasets
- final_published_versions
- rights_documentation

bwatson78:
@eporter23 `grant_information_notes` is already established within the CurateGenericWork model as `grant_information`. Please use `grant_information` as the header in the csv.

bwatson78:
There is no UI example I can display.

https://github.com/emory-libraries/dlp-curate/pull/1320

eporter23:
@bwatson78, impressive work here! I started testing and ran into an error when I input a Creative Commons license value (from our locally defined vocabulary) in URL form into the CSV, using the _re_use_license_ header:  
http://creativecommons.org/licenses/by/3.0/us/ 

This URL was copied directly [from our term list here](https://github.com/emory-libraries/dlp-curate/blob/main/config/authorities/licenses.yml). Should I be using the string value instead?

See the error below. Note that I had the following also:
http://rightsstatements.org/vocab/NoC-US/1.0/ assigned to _rights_statement_ so I assume that error message was actually referring to the re_use_license entry.

![Screen Shot 2020-07-14 at 4.22.04 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/eada1128-c4b2-43bb-babb-ca907a607e92)

bwatson78:
@eporter23 You’re right, it’s on the list, but the mapper method is using the same logic as rights_statement, which is only focused on licenses marked active. I do have to change the error statement to reflect the right verbiage. Below are the licenses I have:

‘’’
terms:
[...]

PR #1320, opened 2020-07-07
Maps additional fields to the import process (#1265).

bwatson78:
- app/importers/curate_mapper.rb: adds requested fields to the CurateMapper that handles the fields read by the CSV importer.
- spec/fixtures/csv_import/good/langmuir_post_processing.csv: injects fields and values into csv so that an import process can be properly tested.
- spec/importers/curate_mapper_spec.rb: inserts tests for the new fields introduced.
- spec/system/importer_preprocessed_langmuir_spec.rb: sets expectations that new fields will be saved.


bwatson78:
@devanshu-m Thank you for suggesting I test further. I found some things that would cause failures in production.

app/importers/curate_mapper.rb
```diff
@@ -4,51 +4,80 @@ class CurateMapper < Zizia::HashMapper
...
+    parent_title:                 "parent_title",
+    pcdm_use:                     "pcdm_use",
+    place_of_production:          "place_of_production",
+    primary_language:             "primary_language",
+    primary_repository_ID:        "primary_repository_ID",
+    publisher:                    "publisher",
+    related_datasets:             "related_datasets",
+    related_material_notes:       "related_material_notes",
+    related_publications:         "related_publications",
+    re_use_license:               "re_use_license",
```
devanshu-m, line 56: More of a question than change: Will re_use_license field here know to use the method later defined in this file? Asking because mapped value is in quotes.

bwatson78: Yes it does.

bwatson78: The testing shows that it does hit the method and produce an RuntimeError when it doesn't have the right url.

PR #1324, opened 2020-07-08
Add rights status for parent works

wulman16:
- "Copyright Varies by Work" is now available as a rights status for parent works whose child works do not have the same rights status.
- Instead of a URL, the value for this field is "Not Applicable." Lux will check for the presence of this value and conditionally render the rights status as a link based on whether it is a URL or not.


wulman16:
Screen Shot 2020-07-07 at 8 55 12 PM

Issue #1290, opened 2020-06-24
Investigate/test Lux UV configuration in Curate


eporter23:
As a pre-cursor to implementation (ticket #1088), some investigation may be needed. 

Based on our understanding of how the access control behavior should function:
1. Authorization to see a work occurs through Curate user group permissions
2. The file resolution and download options in the UV are triggered based on the Visibility assigned to a given work
3. The UV configuration and resolution that is shown to users is _always the same_ for a work, regardless of the user level. A user can either see the work (and its UV) or cannot: we do not show different resolution or download options based on a user's credentials. 

In this ticket, we should first research what was implemented for the Lux UV and identify changes to make in Curate.

- [ ] Confirm changes made to UV configuration for Lux (review commits, etc.)
- [ ] Confirm how the dynamic UV configuration switching occurs in Lux and what needs to be added for Curate
- [ ] Test the Lux UV configuration in curate-arch only
- [ ] Identify any gaps or issues that arise before full implementation

eporter23:
@bwatson78 FYI, we wrote up a prequel ticket for #1088 and this documents our discussion in sprint planning on 6/24. This ticket also clarifies expected behavior for how the access controls function.

bwatson78:
@eporter23 I researched the way that Lux UV permissions are coded and have come up with this summary:

The universal viewer has **three configurations**.
- **Standard Config**: Download, Fullscreen, and Share disabled; Paging Toggle Enabled;
- **Liberal Config**: Download, Fullscreen, and Share enabled; Paging Toggle Enabled;
- **Liberal Low Config**: Fullscreen, and Share enabled; Paging Toggle Enabled; Download Dialogue: Current View and Low Res Option Hidden, whole image restricted to 400px;

**Visibility**
- **"open" OR ("authenticated" AND user_signed_in? true)** = Liberal Config
- **"emory_low" AND user_signed_in? true** = Liberal Low Config
- **Anything Else** = Standard Config

**Conclusion**
- All users see the UV and at least Paging Toggle feature
- If the work is marked "open" or "authenticated" with a user signed in, everything is available
- If a user is signed in but the work is marked "emory_low", then all features are available, but the image is restricted to 400px max.

**Rose Room Notes**
- Since all of the works denoted to the Rose Room are only able to be delivered in the said room, the Universal Viewer doesn't have any restrictions directly related to the room.

My analysis on how to proceed is to follow.

bwatson78:
[...]

PR #1339, opened 2020-07-15
Matches the configuration of Curate UV to Lux (#1290).

bwatson78:
- app/assets/javascripts/uv_width.js: resizes the viewer to the container width.
- app/assets/stylesheets/curate.scss: styling that enforces the viewer to exist in desktop mode.
- app/controllers/application_controller.rb: pastes and slightly alters the actions from Lux' setup.
- app/helpers/hyrax/iiif_helper.rb: points the configuration to the uv route.
- app/values/uv_configuration.rb: imports model from Lux for configuration generation. 
- app/views/hyrax/base/iiif_viewers/_universal_viewer.html.erb: changes size and helper call.
- config/routes.rb: adds new route for json action.
- config/uv/uv-config.json and config/uv/uv_config_liberal.json: brings over the same code from Lux, even though this seems to only be there for redundancy.
- config/uv/uv.html: mirrors the Lux file.
- package.json: copies Lux' PR.
- spec/*: uses the same spec as Lux with only minor alterations.

### DIFF
--- a/app/assets/javascripts/uv_width.js
+++ b/app/assets/javascripts/uv_width.js
@@ -1,0 +1,7 @@
+$(document).on('turbolinks:load', function() {
+  $('#universal-viewer-iframe').width($('.view-wrapper').width())
+
+  $(window).on('resize', function(){
+    $('#universal-viewer-iframe').width($('.view-wrapper').width())
+  })
+})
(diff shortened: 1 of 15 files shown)


bwatson78:
Each image is displaying its visibility level.

Screen Shot 2020-07-15 at 1 26 28 PM

Screen Shot 2020-07-15 at 1 26 51 PM

Screen Shot 2020-07-15 at 1 38 05 PM

Issue #1190, opened 2020-05-20
Configure facets for Curate


eporter23:
The Curate application currently has some default/automatically generated facets displaying from Hyrax.

We need to customize the Curate facets so they closely replicate the Lux facets.

- [x] Library
- [x] Collection
- [x] Creator
- [x] Format
- [x] Genre
- [x] Language
- [x] Date* (see notes below)
- [x] Publication Date** (new facet on _date_issued_ or _human_readable_date_issued_ ?)
- [x] Subject - Topics
- [x] Subject - Names
- [x] Subject - Geographic
- [x] Rights Status
- [x] Access

For the Date facet, we can either try to: 
- replicate the same date slider as Lux offers, OR
- add a Date facet that shows a list of years (using the normalized "date" value used by the Lux date slider)

Date Published will be a new facet that only includes publication dates (_date_issued_ or _date_issued_ or _human_readable_date_issued_ ). This will be helpful for investigating copyright-related questions.
- This could either use the Lux date slider, or a list of dates. Typically _date_issued_ is provided as a year only, but this could vary as we ingest more material.

bwatson78:
@eporter23 The process of getting the fields to show correctly is simple, but getting the slider to work, at least on my local machine, is proving to be quite difficult. The slider function relies heavily on the `@response` from solr having the `stats` attribute populated, and that's not the case. I have been attempting to hack the gem with overrides, but for every piece of the gem I fix, another broken element pops up. 

At this point, It may be better to opt for the years list based on the human readable year elements. If you wanted me to estimate how long it would take to get everything working on the slider, I'd say Monday or Tuesday.

eporter23:
@bwatson78 I'm happy to hear things are largely working! If we simply have the years list, is there any way to limit to a range of multiple dates, or can the user only select one at a time? With the date searches that Library Staff will want to do, one expected use case is to try to limit for "dates between x and y" or "dates before x". We can also look at the slider as a future enhancement.

bwatson78:
@eporter23 If you're alright with a Tuesday delivery, the from-to was populating with that gem. I tried using it and it wasn't narrowing the works as expected. I think I worked on the same issue in Lux, so I can circle back there and see what I did to fix that. Just let me know if you want to move forward with that and I'll jump on it.

eporter23:
@bwatson78 sure, Tuesday is fine and if it rolls over into the next sprint that's okay too because you're ahead of schedule with this one!

[...]

PR #1343, opened 2020-07-16
Swaps out human-readable fields with year isims (#1190).

bwatson78:


bwatson78:
Screen Shot 2020-07-16 at 11 09 59 AM

Issue #1191, opened 2020-04-24
Edit display of Rights Status in UV


nikdragovic:
As an end user, I would like to view a clickable link with human-readable text in the Rights Status field, so that it is easy to ascertain the rights status from the embedded viewer and easy to click for more information about that status.

## Acceptance Criteria
- [x] Instead of static text in the Rights Status Field, add linked text that replicates what displays in the Rights Status field in the Access and Copyright section.
- [x] Clicking the link should open the target in a new tab, as it does for PURL.

![Screen Shot 2020-05-20 at 8.56.49 AM.png](https://images.zenhubusercontent.com/5ba135cf945aef49576d50f2/ccd9d126-555c-4504-b301-559b642ffa36)

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1351

bwatson78:
![Screen Shot 2020-07-20 at 12.30.44 PM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/ab45a68a-8b6b-42cd-868b-84e527d2ad17)

bwatson78:
For my local environment, I had to modify my object before the Manifest Builder service would recreate the metadata since the builder only creates a new one if the date modified differs the manifest file key.

bwatson78:
Since this delivers the same UV to both Lux and Curate, there shouldn't be any alterations necessary for the Lux side.

eporter23:
Looks good! Tested by manually regenerating the manifests on these objects:
https://curate-test.library.emory.edu/concern/curate_generic_works/163stqjqb4-cor?locale=en
https://curate-test.library.emory.edu/concern/curate_generic_works/7440rxwdnt-cor?locale=en

We'll plan to do a large-scale regeneration of manifests in production sometime soon to pick up this change environment-wide.

eporter23:
Update: the manifest regeneration process is still running, but some works are now showing the updated rights status display.

PR #1351, opened 2020-07-20
Changes Rights Status to a link with QA terms as text (#1191).

bwatson78:
- app/presenters/hyrax/curate_generic_work_presenter.rb: swaps out just displaying value with formatting a link similar to "Persistent URL".
- spec/*: changes expectations for the link processing.


bwatson78:
Screen Shot 2020-07-20 at 12 30 44 PM

Issue #1252, opened 2020-06-11
Provide a count of files ingested per Collection


eporter23:
As part of migration acceptance criteria, we need a way to count the number of _all individual files_ (not just FileSets) ingested for each Collection. This is not currently available in the Hyrax UI, which provides an overall count of FileSets only, and also does not report FileSet counts at the Collection level.

In SOLR, file-level information is also not indexed. From SOLR, is possible to get a count of files per fileset, and a list of the filesets attached to each work, and a list of the works assigned to each Collection but this data would have to collated. 

Ideas discussed:

1. Create a background job to count all works in a collection, then find its filesets, then count files per fileset

- User supplies a collection ID
- Job outputs a CSV or JSON file that contains:
- Total count of filesets in collection
- Total count of all files in collection

eporter23:
Possible approach: background job to count all works in a collection, then find its filesets, then count files per fileset

bwatson78:
Blocked: awaiting instructions from @devanshu-m on how to deliver the file to an AWS bucket.

bwatson78:
@eporter23 After speaking with Devanshu, who, earlier, discussed with Collin about setting up a connection to AWS for delivery, he let me know that we should put off that delivery system setup until we create the UI pages for the rake tasks. The only interaction that we have currently with aws buckets is the checking of binaries against our sha1 values--nothing pushing to a separate bucket.

So, for the time being, this would be processed the same way that we do preprocessing tickets.

With that in mind, once we get this into test/arch, please send me a ticket to request the resulting json for this process.

eporter23:
@bwatson78 that makes sense, thanks for the update.

bwatson78:
No screenshot to display. Will need to be tested once deployed to a production environment.

bwatson78:
Produced error on test run. Will debug.

bwatson78:
Pr for fix: https://github.com/emory-libraries/dlp-curate/pull/1363

eporter23:
@bwatson78 when you're back in the office, can we coordinate on running this in prod?

bwatson78:
@eporter23 New Job d414229f60675facac8ec2e5 kicked off at 2020-08-11T15:25:50Z

bwatson78:
File delivered to @eporter23 via slack.

eporter23:
Looks good, thanks!

PR #1353, opened 2020-07-20
Exports Collection file statistics to a json file (#1252).

bwatson78:
- app/jobs/collection_files_ingested_job.rb: pulls and writes file stats to a json file.
- lib/tasks/collection_files_ingested.rake: creates rake task to process the above job.
- spec/*: ensures above job works correctly.

### DIFF
--- a/app/jobs/collection_files_ingested_job.rb
+++ b/app/jobs/collection_files_ingested_job.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+class CollectionFilesIngestedJob < Hyrax::ApplicationJob
+  def perform(collection_array = nil)
+    collections = collection_array.present? ? selected_coll_hsh(collection_array) : collection_hsh
+
+    collection_hashes = collections.map do |hsh|
+      works = collection_works(hsh[:id])
+      filesets = filesets(fileset_ids(works))
+
+      {
+        "collection_id" =>    hsh[:id],
+        "collection_title" => hsh[:title]&.first,
+        "work_total" =>       works.size,
+        "fileset_total" =>    filesets.size,
+        "file_total" =>       filesets_file_count(filesets)
+      }
+    end
+    File.open(Rails.root.join('public', "collection-counts-for-#{Time.now.strftime("%Y%m%dT%H%M")}.json"), "w") do |f|
+      f.write(collection_hashes.to_json)
+    end
+  end
+
+  private
+
+  def solr_service
+    Hyrax::SolrService
+  end
+
+  def query_builder
+    Hyrax::SolrQueryBuilderService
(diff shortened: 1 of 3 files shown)


app/jobs/collection_files_ingested_job.rb
```diff
@@ -0,0 +1,59 @@
...
+        ), rows: 1_000_000
+      )
+    end
+
+    def fileset_ids(works)
+      works&.map { |w| w['file_set_ids_ssim'] }&.compact&.flatten
+    end
+
+    def filesets(ids)
+      ids.map { |id| FileSet.find(id)&.to_solr }
```
devanshu-m, line 53: Instead of calling to_solr on a file_set, do you think it might be a better idea to retrieve the already indexed solr file_set object since you are using solr_service already somewhere in the file?

lib/tasks/collection_files_ingested.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+namespace :curate do
+  namespace :collection_files_ingested do
+    desc 'Process delivery of ingested files count json file'
+    # If passing along multiple ids, they must be separated by one space each.
+    task process: :environment do
+      collection_array = ENV['collections']&.split(' ')
+
+      CollectionFilesIngestedJob.perform_now(collection_array)
```
devanshu-m, line 9: Do you think it'll be a better idea to have this as a background job instead of inline? Since our collections are large, there are chances of broken pipes when running rake tasks that kick off jobs.

spec/jobs/collection_files_ingested_job_spec.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CollectionFilesIngestedJob, :clean, perform_enqueued: [AttachFilesToWorkJob] do
```
devanshu-m, line 5: I don't think you need to enqueue this job here. Also, can you add one more file to the file_set so we can check the file count is 2?

app/jobs/collection_files_ingested_job.rb
bwatson78: I will fix this today.

bwatson78: Change made.

lib/tasks/collection_files_ingested.rake
bwatson78: Change made.

spec/jobs/collection_files_ingested_job_spec.rb
bwatson78: Change made.

Issue #1264, opened 2020-06-18
Post-ingest preservation events - log as the initiating user, not depositor


eporter23:
For manually initiated post-ingest preservation events such as fixity checking an individual fileset in the UI, we should ensure that the event is recorded as having been performed by the actual system user who initiated the action. It currently appears to default to the depositing user.

Tested in Curate-test:

Kathryn performed a fixity check on this fileset, which had no prior fixity check events, and which was part of a work that I deposited:
https://curate-test.library.emory.edu/concern/parent/187pnvx0tt-cor/file_sets/139x0k6dsk-cor

The fixity check preservation events were generated with my name as the user, but I am not the one who executed them.

![Screen Shot 2020-06-18 at 3.29.58 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/8dcfdb73-5c61-4997-938a-7693bf3c5cc0)

eporter23:
TBD: for the fixity check event specifically, do we need to confirm who the user is when we start running the rolling fixity check job?

eporter23:
TBD: is this a Hyrax bug?

bwatson78:
@eporter23 Inherently, the Hyrax-inherited process isn't aware of the initiating user, since it's passed from the controller and then through two more classes without the the current user being passed along to them. This can be fixed by just making the current user an argument that the two classes down the chain can expect and receive. The only problem arises when the Fixity Checks are performed via rake task. How do you want that be addressed?

It is just a string it takes, but I'm wondering how automated versus informative you want the string to be. We could set it to just `software_engineering`, but that could be one of five of us. I could make it so the rake task takes in a string value for an engineer (something like `bwatson`), but if someone other than you or Kathryn is working on the system, they may not be aware of who we, in software engineering, are, which may not help you much. It's your choice.

eporter23:
@bwatson78 for a UI-based fixity check, it should be the user who clicked the button (e.g. _eporter_). For a rake task, let's set the user to _Curate system_ which will be more meaningful for the long haul of preservation audit trails.

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1353

eporter23:
Looks good in prod!

PR #1357, opened 2020-07-21
Replaces the depositor of the file with the current_user uid (#1264).

bwatson78:
- app/controllers/hyrax/fixity_checks_controller.rb: includes the current user's `uid` as `initiating_user` argument.
- app/jobs/fixity_check_job.rb: brings in the new `initiating_user` argument and uses it instead of `fixity_file_set.depositor`.
- app/services/hyrax/file_set_fixity_check_service.rb: overrides the hyrax service to accept the new argument.
- lib/tasks/fixity_check.rake: hardcodes the preferred user value into the rake call associated.
- spec/*: inserts the expectation for the `initiating_user` argument.

### DIFF
--- a/app/controllers/hyrax/fixity_checks_controller.rb
+++ b/app/controllers/hyrax/fixity_checks_controller.rb
@@ -31,7 +31,7 @@ module Hyrax
         # have to sometimes return a 'in progress' status for some bytestreams,
         # which is a possible future enhancement.
         @fixity_check_service ||=
-          FileSetFixityCheckService.new(::FileSet.find(params[:file_set_id]), async_jobs: false)
+          FileSetFixityCheckService.new(::FileSet.find(params[:file_set_id]), async_jobs: false, initiating_user: current_user.uid)
       end
   end
 end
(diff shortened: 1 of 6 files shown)


devanshu-m:
Instead of making changes to the job, service, and other places, do you think we can just check if a current_user exists and if not (then we assume fixity checking is through the rake task) assign it as Curate System? Only change we would need to do to achieve this would be here: https://github.com/emory-libraries/dlp-curate/pull/1357/files#diff-dfcc26dfd2d5430d81c3ed60468ba9e4R74

bwatson78:
@devanshu-m If you're asking if `current_user` exists in the job or service, it doesn't. Trust me, I tried it with byebug and it's empty. I'll try it again to make sure. Give me a few moments.

bwatson78:
```(byebug) current_user
*** NameError Exception: undefined local variable or method `current_user' for #<Hyrax::FileSetFixityCheckService:0x00007ff7aa8c44d0>

nil
(byebug) session
*** NameError Exception: undefined local variable or method `session' for #<Hyrax::FileSetFixityCheckService:0x00007ff7aa8c44d0>

nil
```

bwatson78:
@devanshu-m I wish it was as easy as that, but `current_user` isn't available past the controller unless you pass it as an argument.

devanshu-m:
Ok, can you copy over specs for the files that are overwritten here from hyrax and make sure tests are passing?

bwatson78:
@devanshu-m RSpec has been imported and updated.

PR #1365, opened 2020-07-28
Formats the banner path agnostically (#672).

bwatson78:
- app/indexers/curate_collection_indexer.rb: formats path without caring what comes before `/branding`.
- spec/indexers/collection_indexer_spec.rb: checks that formatting is working using a path we have in `test` envirnoment.

### DIFF
--- a/app/indexers/curate_collection_indexer.rb
+++ b/app/indexers/curate_collection_indexer.rb
@@ -30,12 +30,7 @@ class CurateCollectionIndexer < Hyrax::CollectionIndexer
   end
 
   def path_sanitized(path)
-    if ENV['BRANDING_PATH'].present? && path.include?(ENV['BRANDING_PATH'].to_s)
-      path.gsub ENV['BRANDING_PATH'], '/branding'
-    elsif path.include? Rails.root.join('public', 'branding').to_s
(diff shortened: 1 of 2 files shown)


app/indexers/curate_collection_indexer.rb
```diff
@@ -30,12 +30,7 @@ def branding_details
...
   def path_sanitized(path)
-    if ENV['BRANDING_PATH'].present? && path.include?(ENV['BRANDING_PATH'].to_s)
-      path.gsub ENV['BRANDING_PATH'], '/branding'
-    elsif path.include? Rails.root.join('public', 'branding').to_s
-      path.gsub Rails.root.join('public').to_s, ''
-    else
-      path
-    end
+    return '/branding' + path.split('/branding').last if path&.include? '/branding'
+    path
```
alexBLR, line 34: you don't need **path** here if you already use return in the method

bwatson78: Oh, I intended that as an `if/else` type of statements, but shorter. The `return` line only fires off if the `path` has the branding string in it (cause it should.) So if the `path` variable doesn't, it just returns the same argument without formatting it.

Issue #1347, opened 2020-07-17
Adjust Sensitive Material field values in work form


eporter23:
The form currently shows true/false as selectable values, but the [original terms specified for this field ](https://docs.google.com/spreadsheets/d/1Z0ebRLaPa4bEJayvTiqwU4ulFglWrpQPgVX6r6v29uY/edit#gid=714216589&range=O60) were "Yes" or "No".

Current output

![Screen Shot 2020-07-17 at 10.31.34 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/dd67f796-acfc-4c9e-8ad4-10d2a5bf9d1f)

Desired output

![Screen Shot 2020-07-17 at 10.32.28 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/4f82e6f3-ebd6-4891-961c-fa639cad53fb)

Alternate output

Make this a simple checkbox with a value of _Yes_
![Screen Shot 2020-07-17 at 10.34.07 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/74ea1fb5-170c-4b91-bda1-26b18394b130)

eporter23:
Recommend going with Yes / No in the drop-down in case we have a local vocabulary defined as true/false.

bwatson78:
![Screen Shot 2020-07-30 at 9.50.02 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/aee697fb-88b8-47b9-98da-7689b66eae8d)
![Screen Shot 2020-07-30 at 9.51.37 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/ef3a8b83-0495-422c-9bca-2cdd42260783)

PR made: https://github.com/emory-libraries/dlp-curate/pull/1373

eporter23:
Form looks good - I'll need to do a little more testing on works that have this value populated already, and with the CSV import process.

eporter23:
I spot checked this on works that already have the value populated and it seems to be translating fine.

eporter23:
Spot checked some sample works using this value in prod as well, looks good.

PR #1373, opened 2020-07-30
Changes sensitive material labels to Yes and No (#1347).

bwatson78:
- app/services/hyrax/sensitive_material_service.rb: overrides the QA method so that it delivers Yes and No.
- spec/system/create_curate_generic_work_spec.rb: tests that the labels are correct.


bwatson78:
Screen Shot 2020-07-30 at 9 50 02 AM
Screen Shot 2020-07-30 at 9 51 37 AM

Issue #1345, opened 2020-07-16
Configure or hide Contact form in Curate


eporter23:
The Curate main navigation shows a link to a Contact page, which contains a form that is not configured.

https://curate-test.library.emory.edu/contact?locale=en

Options:

1. Remove the link to this form
2. Configure the form so that it emails to a valid address: dlpfeedback@listserv.cc.emory.edu
3. **Redirect the Contact link so that it redirects to https://emory.libwizard.com/f/dlp-feedback**

eporter23:
Discussed: overwrite the route to the Contact page so that it redirects to our DLP Feedback form. This could then be updated to redirect to service now (or final support url).

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1374
![Screen Shot 2020-07-30 at 8.06.56 PM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/6f2f9070-0120-497f-8e52-bcac9b0ce2aa)
![Screen Shot 2020-07-30 at 8.06.47 PM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/70914e8a-3ad9-4357-8df2-d8284ec79ae6)

eporter23:
Looks good!

eporter23:
Verified in prod.

PR #1374, opened 2020-07-31
Changes contact link to libwizard service (#1345).

bwatson78:
- app/views/_controls.html.erb: brought in to override hyrax' link.
- spec/system/viewing_homepage_spec.rb: creates test to make sure link goes to the right destination.


bwatson78:
Screen Shot 2020-07-30 at 8 06 56 PM
Screen Shot 2020-07-30 at 8 06 47 PM

Issue #1267, opened 2020-06-19
Make Collection View pages metadata easier to read


eporter23:
Collection View pages' metadata display uses default Hyrax styles, which are not optimal for reading. 

[Mockup for layout revisions](https://docs.google.com/document/d/19aduQc_g6drxWDtf3szQnUTSK_XA5EG4MynOVZhmKYU/edit#)

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1375
![Screen Shot 2020-08-03 at 9.54.41 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/c8542da3-cb13-457a-a35b-39211956b7af)
![Screen Shot 2020-08-03 at 9.53.59 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/e6ea7120-77a8-46fb-a8fd-84acfb6b1d8b)

eporter23:
Looks great @bwatson78!

eporter23:
Such a huge improvement - thanks!

PR #1375, opened 2020-08-03
Styles the collection view pages and removes size from listed metadata (#1267).

bwatson78:
- app/assets/stylesheets/curate.scss: adjusts the styling of the elements addressed in the ticket.
- app/presenters/hyrax/collection_presenter.rb: removes the display value of size, per request by @eporter23 made on 7/31/20 via DM on slack.
- app/views/hyrax/collections/_show_descriptions.html.erb: strips `dl` structure in favor of a straight bootstrap column approach.
- app/views/hyrax/collections/show.html.erb: brings in the page as override to excise exact column sizing.
- app/views/hyrax/dashboard/collections/_show_descriptions.html.erb: once again removes dl structure fo that label-top/value-bottom is maintained.

### DIFF
--- a/app/assets/stylesheets/curate.scss
+++ b/app/assets/stylesheets/curate.scss
@@ -54,8 +54,33 @@ dl.file-show-details {
   }
 }
 
+div.hyc-metadata {
+  dl {
+    border-top: none;
+  }
+
+  div {
+    border-bottom: none;
+  }
+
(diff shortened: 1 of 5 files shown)


bwatson78:
Screen Shot 2020-08-03 at 9 54 41 AM
Screen Shot 2020-08-03 at 9 53 59 AM

Issue #1346, opened 2020-07-16
Adjust field display labels on search result snippets


eporter23:
In Curate search results, the field labels seem to include suffixes (from SOLR?) which make the results look odd. 

Current output
![Screen Shot 2020-07-16 at 3.30.52 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/455fa696-3fc7-4232-848e-45f5c365cd45)

Desired output - list the following Hyrax and Curate fields using their Display Labels:

- [x] Title (title_tesim) and thumbnail as they currently display
- [x] Library (holding_repository_tesim)
- [x] Collection (member_of_collections_ssim)
- [x] Date Uploaded (date_uploaded_dtsi)
- [x] Date Modified
- [x] Visibility (human_readable_visibility_ssi?)

bwatson78:
@eporter23 I just wanted to confirm that the above list contains the only fields you want to be present in the search list item metadata field. I ask because the current list could display up to 21 fields, and we're bringing the list down to 5, if so.

bwatson78:
Fields currently assigned to index:

- title
- description
- keyword
- subject
- creator
- contributors
- proxy_depositor
- depositor
- publisher
- based_near_label
- language
- date_uploaded
- date_modified
- date_created
- rights_statement
- license
- resource_type
- file_format
- identifier
- embargo_release_date
- lease_expiration_date

eporter23:
@bwatson78 thanks, that is very helpful. Based on the list you provided, I'd like to keep the smaller set indicated in the description above. I did add Date Modified also.

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1377
![Screen Shot 2020-08-03 at 10.43.16 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/a6b9202d-c43c-483f-abfb-2e90cfe941f4)

eporter23:
Looks good!

eporter23:
Tested in prod, looks good!

PR #1377, opened 2020-08-03
Changes the fields show on the search result list (#1346).

bwatson78:
- app/controllers/catalog_controller.rb: removes unwanted fields, adds requested ones, and tailors all of their labels.
- spec/system/search_results_page_spec.rb: tests all fields that are populating.


bwatson78:
Screen Shot 2020-08-03 at 10 43 16 AM

Issue #915, opened 2020-01-26
Date digitized repeated in bulk imported books


eporter23:
To confirm: are MARC date formats always consistent?

The Date Digitized field is repeated multiple times in book objects as shown below.

We also need to correct the date so that it is formatted in EDTF time, because it appears to cause problems with validation when you try to save the work using the work form. The field is not required, but it does show validation errors if you scroll down to view the field.

Ideas to resolve:

- [ ] Change Date Digitized mapping in the books pre-processor so that it pulls from the CSV and not Alma XML
- [ ] OR - temporarily remove Date Digitized in the books pre-processor and we add it manually to CSVs
- [x] OR - refine the logic in the pre-processor so that it only uses the first date, and trims that to a 4 digit year

![Screen Shot 2020-01-26 at 11.58.42 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/3b6df3a6-a489-4932-96a0-973e0eaa4f03)

bwatson78:
@eporter23 Should a recurring process be written to check, across all documents, this field to see if it has redundant values, format the unique values into EDTF, persist the repetitive, unformatted values with the corrected values, and maybe reindex the object? 

Or, should we leave the Fedora object value as-is and just adjust how the solr indexing digests and formats the return field?

bwatson78:
@eporter23 Can you also provide me an example object?

eporter23:
@bwatson78 to your first question, Kathryn and I have been manually cleaning up the generated output in the CSVs prior to ingest and/or cleaning up the entries in the objects in Curate. So, there should not be a need to look for these on a recurring basis. So, the goal would be to adjust the output of this that comes out of the preprocessor scripts so we don't have to fix them manually. I'll look for a sample CSV that hasn't been fixed yet.

eporter23:
Here's a small CSV generated by the preprocessor that contains the repeated entries in date_digitized: https://github.com/emory-libraries/cor-data-transfer/blob/master/curate-ingests/2019/processed/yearbooks/Yearbooks-Oxf-1988-and-90.csv

Alma source XML file: https://github.com/emory-libraries/cor-data-transfer/blob/master/curate-ingests/2019/in_development/yearbooks/alma-yearbooks.xml 
See the 583, subfield c entries:
**<datafield ind1=" " ind2=" " tag="583">**
    <subfield code="a">digitized</subfield>
    **<subfield code="c">2010</subfield>**
[...]

PR #1381, opened 2020-08-05
Updates yellowback preprocessor with `date_digitized` adjustment (#915).

bwatson78:
- app/importers/yellowback_preprocessor.rb: adds manipulation logic to import dates correctly.
- spec/fixtures/csv_import/yellowbacks/yellowbacks_marc.xml: alters date so that new preprocessor method can produce an example.
- spec/importers/books_preprocessor_metadata_spec.rb: inserts context concerning how date should be converted to 4-digit year.
- spec/models/curate_generic_work_spec.rb and spec/importers/curate_mapper_spec.rb: makes these tests reflect the expected standard date for `date_digitized` from here on out.


bwatson78:
No screenshots to provide.

PR #1386, opened 2020-08-06
creates ui and functionality for background jobs in admin dashboard

### LINKED ISSUE #1319: UI for bulk-loading preservation workflow metadata
- [x] Curate administrators can load preservation workflow metadata by submitting a CSV in the UI
- [x] the work should happen in a background job

[Wireframe](https://docs.google.com/drawings/d/1I7pLwQFjB9Wlcc2-grcS0Ug04IymOqPXEySrxftoJqI/edit), see also #1318

[[needs details: what info is in the csv?]]

[Sample CSV that is run in the current rake task](https://github.com/emory-libraries/cor-data-transfer/blob/master/curate-ingests/2019/processed/medical-artifacts/medical-artifacts-curate-processed-pwf.csv)

alexBLR:
- Curate administrators can load preservation workflow metadata by submitting a CSV in the UI

- The work should happen in a background job

Resolves #1319

### DIFF
--- a/app/controllers/background_jobs_controller.rb
+++ b/app/controllers/background_jobs_controller.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+class BackgroundJobsController < ApplicationController
+  with_themed_layout 'dashboard'
+  def new; end
+
+  def create
+    if params[:jobs] == 'cleanup'
+      FileSetCleanUpJob.perform_later
+      msg = "File Set Cleanup Job"
+    elsif params[:jobs] == 'preservation'
+      PreservationWorkflowImporterJob.perform_later(params[:preservation_csv].path)
+      msg = "Preservation Workflow Importer Job"
+    else
+      CSV.foreach(params[:reindex_csv].path, headers: true) do |row|
+        r = row.to_h
+        ReindexObjectJob.perform_later(r['id'])
+      end
+      msg = "Reindex Files Job"
+    end
+    flash_message = msg + " has started successfully."
+    redirect_to new_background_job_path, notice: flash_message
+  end
+end
--- a/app/views/hyrax/dashboard/sidebar/_tasks.html.erb
+++ b/app/views/hyrax/dashboard/sidebar/_tasks.html.erb
@@ -23,6 +23,10 @@
   <%= menu.nav_link('/csv_imports/new',  data: { turbolinks: false }) do %>
     <span class="fa fa-upload"></span> <span class="sidebar-action-text">Import Content From a CSV</span>
   <% end %>
+
+  <%= menu.nav_link('/background_jobs/new',  data: { turbolinks: false }) do %>
+    <span class="fa fa-tasks"></span> <span class="sidebar-action-text">Background Jobs</span>
+  <% end %>
 <% end %>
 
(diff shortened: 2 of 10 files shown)


app/controllers/background_jobs_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  with_themed_layout 'dashboard'
+  def new; end
+
+  def create
+    if params[:jobs] == 'cleanup'
+      FilesetCleanupService.new.execute
+    elsif params[:jobs] == 'preservation'
+      LoadPreservationMetadataService.new({ file: params[:preservation_csv] }).execute
+    else
+      ReindexFilesService.new({ file: params[:reindex_csv] }).execute
```
rotated8, line 12: Rather than using a service, can we enqueue these jobs directly? What do you like about using the services?

config/routes.rb
```diff
@@ -62,6 +62,8 @@
     end
   end
 
+  resources :background_jobs
```
rotated8, line 65: As I understand it, `resources` means it creates paths for all the CRUD operations. Is that right? What happens if someone tries to get to background_job#index?

app/views/hyrax/dashboard/sidebar/_tasks.html.erb
```diff
@@ -23,6 +23,10 @@
   <%= menu.nav_link('/csv_imports/new',  data: { turbolinks: false }) do %>
     <span class="fa fa-upload"></span> <span class="sidebar-action-text">Import Content From a CSV</span>
   <% end %>
+
+  <%= menu.nav_link('/background_jobs/new',  data: { turbolinks: false }) do %>
+    <span class="fa fa-tasks"></span> <span class="sidebar-action-text">Background Jobs</span>
+  <% end %>
```
rotated8, line 29: Nice!

spec/services/reindex_files_service_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+RSpec.describe ReindexFilesService do
+  let(:csv_file) do
+    File.open(fixture_path + '/reindex_files.csv')
+  end
+
+  describe 'execute' do
+    it 'loads preservation preservation metadata' do
+      described_class.new({ file: csv_file }).execute
+      expect(ReindexObjectJob).to have_been_enqueued.exactly(:twice)
```
rotated8, line 12: Twice, because there are two ids in the CSV, right?

app/controllers/background_jobs_controller.rb
alexBLR: If we ever wanted to reuse it in a rake task to keep it DRY

spec/services/reindex_files_service_spec.rb
alexBLR: Yes. 2 ids in the fixture

config/routes.rb
alexBLR: You are right. I need to limit it to new and create.

alexBLR:
Changes: removed services and service tests, put everything in the controller, wrote additional tests for the controller, limited routes for background jobs controller, added a dynamic flash message on create.

PR #1399, opened 2020-08-10
Adds initial work for collection overload strategy

devanshu-m:
* We are adding a new metadata field that will store the source_collection_id on the work.


devanshu-m:
This is a draft PR.
@eporter23 I have `source_collection_id` as a scalar value here. Is this ok or do we need this field to be multivalued?

eporter23:
@devanshu-m yes, source_collection_id should be single valued. Thanks for confirming.

eporter23:
@devanshu-m do you want me to close this PR?

Issue #1283, opened 2020-06-23
Re-characterize a FileSet using the UI


eporter23:
This functionality would enable a user to regenerate technical metadata for a specific FileSet on demand using the UI.

This would involve cloning the characterization job, emptying the prior technical metadata, and then kicking off the job for that specific fileset.

To confirm:
Can we also add another Characterization Preservation Event entry when this occurs? This should record the name of the user that submitted the job from the UI.

eporter23:
Hey @bwatson78 I saw that you had picked this one up. I had recently moved it out of the Backlog into New Issues because it isn't a high priority. But if you want to take a look at it, should be fine.

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1417
![Screen Shot 2020-08-17 at 12.13.56 PM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/999514ab-158f-4739-85c4-bd6e51ec4176)

devanshu-m:
@eporter23 is this something we are doing? Re-characterization will append data in some metadata fields instead of overwriting.

eporter23:
@devanshu-m thanks for commenting on this and reminding me of that concern. The ticket description as we wrote it notes that it would delete the existing technical metadata and recreate it. But if that isn't possible (@bwatson78) we should hold off on doing this, because we would not want to append new/different technical metadata on top of existing metadata for files.

bwatson78:
@eporter23 and @devanshu-m: Bringing this back into "In Progress" to dig deeper into how to address that requirement.

bwatson78:
@eporter23 The UI and basic controller action has been fleshed out, but Devanshu feels that Software Engineering needs to meet to discuss the best way to perform the re-characterization in the back end. Placing this back in New Issues.

eporter23:
@bwatson78 looks good! One item from the ticket description above regarding the Preservation Event - can it record the username who initiated the recharacterize event? I just recharacterized a fileset that Kathryn created, but it shows her name as running the re-characterization:

![Screen Shot 2020-09-10 at 9.05.14 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/00efa561-defd-4dbc-aeb0-3968d1735283)

bwatson78:
@eporter23 Hmm, that's interesting. Bringing it back into progress to debug this.

bwatson78:
[...]

PR #1417, opened 2020-08-17
Adds button to re-characterize file set (#1283).

bwatson78:
- .rubocop.yml: skips `characterize_job` file for rubocop failure when adding new argument.
- app/controllers/characterization_controller.rb: creates new action that starts a re-characterization by pressing a button on the file_set show page. Please note, I understand the preference that the action be associated with the closest relative controller, but attempting to do so produced errors that I couldn't find a quick fix for, so I left this action separate.
- app/jobs/characterize_job.rb: adds `user` argument in case of the Characterization taking place from a controller action.
- app/jobs/re_characterize_job.rb: institutes an individual job for the re-characterization of the preservation_master_file.
- app/services/re_characterization_service.rb: brought processing over to a service class for ease of testing.
- app/views/hyrax/file_sets/show.html.erb: inserts the button mentioned above as well as refactoring the text into localizations.
- config/locales/hyrax.en.yml: adds in the text from the FileSet show page into this locale file.
- config/routes.rb: initiates new route to Characterization action.
- spec/presenters/hyrax/admin_set_presenter_spec.rb: addresses, what seems to be, a flaky error that is cleared up by cleaning out Fedora after every test.
- spec/*: creates new tests for the above actions.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,6 +76,7 @@ Metrics/CyclomaticComplexity:
         - app/indexers/curate/file_set_indexer.rb
         - app/indexers/curate_generic_work_indexer.rb
         - app/controllers/iiif_controller.rb
+        - app/jobs/characterize_job.rb
 
 Layout/LineLength:
     Exclude:
--- a/app/controllers/characterization_controller.rb
+++ b/app/controllers/characterization_controller.rb
@@ -5,7 +5,7 @@ class CharacterizationController < ApplicationController
 
   def re_characterize
     file_set = FileSet.find(params[:file_set_id])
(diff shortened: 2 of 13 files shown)


bwatson78:
Screen Shot 2020-08-17 at 12 13 56 PM

devanshu-m:
This PR has work that's already in main. Can you please rebase?

devanshu-m:
Waiting on Emily to reply on ticket https://app.zenhub.com/workspaces/digital-library-project-5bf484ae4b5806bc2bf6875b/issues/emory-libraries/dlp-curate/1283

devanshu-m (approved):

Issue #1385, opened 2020-08-06
Add new setting to Collection Type to differentiate between Source, Deposit Collection views


eporter23:
Because all Hyrax collections seem to share a common display template, the collection type model would need a new option added. This should look similar to existing Collection Type options available when you [create a new Collection Type](https://curate-test.library.emory.edu/admin/collection_types?locale=en).

Example:

Deposit-only Collection: (checkbox) 

We can delete and recreate the existing Deposit Collection type in test, so that it would utilize the new property/setting.

Mockup:

![Screen Shot 2020-08-12 at 8.34.26 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/1540ded5-6d1c-4bd5-b368-0b6f8cd6fd6a)

bwatson78:
@eporter23 Just confirming that this is only the addition of a boolean attribute to the `CollectionType` model and that what it's affecting in the system is not in this ticket's scope.

eporter23:
@bwatson78 that is my understanding. Once this is established, I will go recreate the Deposit Collection type. After that, I'll test depositing some works in there and hopefully we'll have the necessary hooks to differentiate between a Source Collection and a deposit-only collection for display purposes as outlined in other tickets.

bwatson78:
@eporter23 Should it default to false/true?

eporter23:
@bwatson78 I think it should default to false (unchecked).

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1424

bwatson78:
![Screen Shot 2020-08-25 at 10.59.02 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/0c2b72d7-422d-4f29-b48d-b352cc186dd6)

bwatson78:
![Screen Shot 2020-08-25 at 11.02.56 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/f5cbac07-8900-4a53-80ae-09f86b572e09)

eporter23:
I created a new Collection Type instance in curate-test and applied the new settings. https://curate-test.library.emory.edu/admin/collection_types/5/edit?locale=en#settings

PR #1424, opened 2020-08-19
Adds Deposit-Only Collection setting to CollectionType (#1385).

bwatson78:


bwatson78:
Screen Shot 2020-08-25 at 10 59 02 AM

bwatson78:
Screen Shot 2020-08-25 at 11 02 56 AM

bwatson78:
@devanshu-m My first commit used the initializer/`class_eval` approach and produced errors because, somehow, the original `prepends` initializer wasn't operating correctly. The errors I was receiving stated that `low_res' was an unknown argument. My theory was, somehow, the amount of initializers was too much for the system and to test that, bring over the actual Hyrax classes I was trying to `class_eval` and just make the changes there. That theory worked.
For the other question, if you check the db migration, not only did I put `default: false`, but I also included `null: false`, which will automatically fill in `false` to all existing CollectionType instances.

app/views/hyrax/admin/collection_types/_form_settings.html.erb
```diff
@@ -0,0 +1,41 @@
+<p><%= t('.instructions') %></p>
```
devanshu-m, line 1: Is this an overwrite? Can you please mention.

bwatson78: Last commit addresses the comment request.

Issue #1414, opened 2020-08-17
Remove or suppress display of Single Use links in View Fileset page


eporter23:
The View Fileset page includes a block for generating a one-time link to a Fileset. Since we are not using Curate as a public facing interface, this element is extraneous and may be confusing for users.

Remove (or suppress display of):

- [x] Single Use Links section and related output

Screenshot - see the elements outlined in red:

![single use links.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/661eb6bf-3534-469b-a7e3-33293a5a2b2f)

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1425
![Screen Shot 2020-08-20 at 10.03.01 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/d5255c67-b2d0-46d6-8dcd-2ec9a1b857cf)

PR #1425, opened 2020-08-20
Removes the Single-Use Links section from FileSet show page (#1414).

bwatson78:
- app/views/hyrax/file_sets/show.html.erb: deletes `single_use_links` render call.
- spec/system/show_file_spec.rb: tests that Single-Use HTML doesn't appear.


bwatson78:
Screen Shot 2020-08-20 at 10 03 01 AM

Issue #1415, opened 2020-08-17
Remove or suppress display of citation feature and social media links in View Work Pages


eporter23:
The View Work page includes a citation builder link which is not configured, followed by a set of social media/sharing icons. Because we are not using Curate as a public-facing discovery tool, and these options are not functional, we should remove these from display.

To remove/suppress:

- [ ] Citations drop-down
- [ ] Social media icon set

Screenshot (see items outlined in red)

![Social and citation items.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/9f25f7fd-77c8-448b-aeec-f62cc050016c)

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1427
![Screen Shot 2020-08-20 at 11.36.50 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/c9f037aa-982d-4132-9506-d667bfbe90d2)

PR #1427, opened 2020-08-20
Removes Citation and Social Media buttons from Works view page (#1415).

bwatson78:
- app/views/hyrax/base/show.html.erb: omits the Social Media and Citation buttons.
- app/views/hyrax/base/show.html.erb: adds tests making sure they're gone.


bwatson78:
Screen Shot 2020-08-20 at 11 36 50 AM

PR #1437, opened 2020-08-24
Changes total_viewable_items to not care about viewability (#1405).

bwatson78:
app/presenters/hyrax/admin_set_presenter.rb: brings in Hyrax code for override. Changes the `total_viewable_items` scope to `total_items`. Would have imported the `spec` file as well, but the file didn't test this method.


bwatson78:
Since the change here doesn't really affect small case situations, there's really no screenshot to provide--results would show better in production.

bwatson78:
To avoid the "red", I'll bring the test over.

Issue #1388, opened 2020-08-06
Update Collection form to add/edit relationships between source and deposit collections


eporter23:
See [wireframe](https://docs.google.com/drawings/d/1w8W1lAFsUV7kKWkJdnzCR1xsJjyYtel4_XP6SPMRlmU/edit) for assigning collection ids at the collection level.

Discussion: it may be best to make these form fields admin-only. It may also be beneficial to detect the type of collection being edited, and only show source OR deposit collection fields as applicable.

eporter23:
Discussed by slack with @bwatson78 - we can use a drop-down select menu instead of the collection lookup feature shown in the wireframe.

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1449
![Screen Shot 2020-08-26 at 11.32.28 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/3b2d7c1d-ef81-4dae-9fbc-653dd3334102)

eporter23:
@bwatson78 I successfully assigned a relationship via UI between a Deposit Collection back to its Source Collection. In testing I also see that the Source Collection displays in one of the collection views but not the "Public" view. Should I create another ticket for this? This also is probably more about ticket #1390.

Deposit Collection - Dashboard View
https://curate-test.library.emory.edu/dashboard/collections/235gb5mkx3-cor?locale=en

Deposit Collection - Hyrax "Public View"
https://curate-test.library.emory.edu/collections/235gb5mkx3-cor?locale=en

eporter23:
@bwatson78 I added the above comments to ticket #1390. The assignment via UI is working as expected between the two collections as shown below:

Deposit Collection
![Screen Shot 2020-08-26 at 7.50.53 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/db3d993a-6321-4d2e-9b29-1f654a717f63)

Source Collection
![Screen Shot 2020-08-26 at 7.51.35 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/b4b3f5df-e251-41c5-ac72-82956ba12742)

bwatson78:
@eporter23 This only reflects on #1390 in the sense that, when the source id is assigned, it would populate if both view pages had the source collection link html. If the form successfully persists the id, then I think the scope of this ticket is complete.

eporter23:
[...]

PR #1449, opened 2020-08-26
Add source and deposit fields to Collection Edit/New form (#1388).

bwatson78:
- app/forms/hyrax/forms/collection_form.rb: adds new Deposit/Source id fields to the Collection Form class.
- app/helpers/form_helper.rb: creates a helper to dry up the Collections' collection.
- app/indexers/curate_collection_indexer.rb: since `deposit_collection_title_ssim` would not work now that the field has been changed to multivalued, had to update it to a tesim field and change the logic.
- app/models/collection.rb: allows multiple values.
- _source_collection_id.html.erb and _deposit_collection_ids.html.erb: fleshes out the the form input fields for the new attributes.
- config/locales/hyrax.en.yml: inserts localizations for both fields for the edit/new labels.
- spec/*: creates or modifies tests related to the changes for both fields and making `deposit_collection_id` multivalued.


app/indexers/curate_collection_indexer.rb
```diff
@@ -13,7 +13,7 @@ def generate_solr_document
       solr_doc['generic_type_sim'] = ["Collection"]
       solr_doc['banner_path_ss'] = banner_path
       solr_doc['source_collection_title_ssim'] = source_collection
-      solr_doc['deposit_collection_title_ssim'] = deposit_collection
+      solr_doc['deposit_collection_titles_tesim'] = deposit_collection&.first
```
devanshu-m, line 16: Why are we calling first on this? Since this is multivalued, there is a possibility of an array being returned with more than one deposit collection.

Suggested change:
`solr_doc['deposit_collection_titles_tesim'] = [deposit_collection]`

app/models/collection.rb
```diff
@@ -38,7 +38,7 @@ def assign_id
     index.as :stored_searchable, :facetable
   end
 
-  property :deposit_collection_id, predicate: "http://metadata.emory.edu/vocab/cor-terms#depositCollectionID", multiple: false do |index|
+  property :deposit_collection_ids, predicate: "http://metadata.emory.edu/vocab/cor-terms#depositCollectionID" do |index|
```
devanshu-m, line 41: Spec for this change is missing. Can you add a unit test in the model?

bwatson78: Yeah, I can

app/indexers/curate_collection_indexer.rb
bwatson78: With just passing the code in, I was getting [["Title"]]. So I needed to eliminate the double array effect.

app/models/collection.rb
bwatson78: @devanshu-m Change addressed.

bwatson78:
Screen Shot 2020-08-26 at 11 32 28 AM

PR #1471, opened 2020-09-04
Corrects errors found when troubleshooting the BackgroundJobsController spec.

bwatson78:
**`spec/controllers/background_jobs_controller_spec.rb`**
1. L# 7, 19-23: have to clear jobs after each test since Jobs were still in queue when next test started, throwing errors.
2. L# 37, 42: csv files were being called in the wrong posts--swapped them out.
3. L# 44: to avoid errors, we need to call out how many times the job will be enqueued.


alexBLR (approved):
👍

PR #1481, opened 2020-09-10
Require source_collection_id on works

wulman16:
- The partial for `source_collection_id` on the new/edit form now checks for whether the form is for a work or a collection. That way, works can require `source_collection_id` without also making it required for collections.


alexBLR (approved):
👍

PR #1483, opened 2020-09-10
Adds manifest regen button

devanshu-m:
* We are adding a manifest regen button for admins to manually regen manifests
on demand.

* **app/controllers/manifest_regeneration_controller.rb**: Adds logic to delete existing manifest for a work if it exists and then call the manifest builder service to create a new manifest.
* **app/views/hyrax/base/_show_actions.html.erb**: Adds button
* **config/routes.rb**: Adds route
* **spec/controllers/manifest_regeneration_controller_spec.rb**: Adds tests

Output:
image


alexBLR (approved):
👍

PR #1488, opened 2020-09-11
Improving manifest key generation in work indexer

devanshu-m:
* There are instances when the order of the rendering_ids in its array gets messed up and we end up with different hash keys whenever a work is reindexed. To be able to have consistency, we are making sure we first sort the array and then pass it for hash generation. We are also checking to make sure holding repo is always single valued (in case the solr field value is used which is an array).


alexBLR (approved):
👍

Issue #1435, opened 2020-08-21
Update rake task documentation for recent changes


eporter23:
- [ ] Manifest regeneration
- [ ] Counts by Collection report
- [ ] Others?

bwatson78:
@devanshu-m I was just checking the changes for manifest generation and, as far as I can tell, removing all of the manifests from the command line is still a necessity when processing all Works' manifests. Am I assuming correctly? If so, the process described in the tutorial is the same.

bwatson78:
@eporter23 Pending @devanshu-m response to the question above, nothing about the tasks themselves have changed. I have changed some descriptions because they are misleading or needed to be expounded upon. A new PR will be available soon.

bwatson78:
New PR made: https://github.com/emory-libraries/dlp-curate/pull/1491

eporter23:
@bwatson78 sounds good, thanks for the updates.

devanshu-m:
@bwatson78 yes, deleting manifests when regenerating for all works is still a necessity.

bwatson78:
That PR was merged and deployed into test/arch.

PR #1491, opened 2020-09-15
Updates the Rake task README with corrected verbiage (#1435).

bwatson78:
`lib/tasks/README.md`:
- L# 25-30: all of the arguments are required.
- L# 41, 44: need to denote Source Collection.
- L# 45: warns user that ids need to come from Source Collections.
- L# 64: since determination of if a file has been modified has changed, have removed the line that refers to the old process.


alexBLR (approved):
👍

PR #1495, opened 2020-09-17
Points the Message Digest Calculation event point to initiating user when processing Recharacterization (#1283).

bwatson78:
- app/jobs/characterize_job.rb: passes the user to the the service if present.
- config/initializers/characterization_service.rb: overrides the `run` and `initialization` methods so they accept a user argument. Also establishes a `@user` instance variable that can be used in the event creation.
- spec/jobs/characterize_job_spec.rb and spec/jobs/re_characterize_job_spec.rb: changes `rspec` expectation for new argument.
- spec/services/characterization_service_spec.rb: checks that initiating user's `uid` is assigned correctly.


bwatson78:
Screen Shot 2020-09-17 at 4 14 45 PM

wulman16 (approved):

Issue #1490, opened 2020-09-15
Show Deposit and Source Collection in main search results for works


eporter23:
The main search results snippet for works shows only the member_of_collection relationship, but ideally should include the Source Collection relationship if different than the Deposit collection:

Current output:
![Coll in search results.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/30a1fade-c62e-4651-ba63-feba4c8f910c)

Desired output:

Use display logic similar to that which is shown on View Works > Relations
If the source_collection_id is the same as member_of_collections_id, show the one Collection title
If the source_collection_id is the different than member_of_collections_id, show both Collection titles

![Screen Shot 2020-09-15 at 9.12.10 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/7ea65e9d-cc35-45c5-a973-bed279181942)

eporter23:
@devanshu-m can you clarify how this has been implemented? I have looked at a few works that are in a Deposit Collection and a Source collection, and if I do a search on them in the main searchbox the results preview shows the Source Collection title only (not both the Deposit and the Source Collection). If that is the expected behavior, that's okay but we may need to change the label to indicate "Source Collection" specifically.

https://curate-test.library.emory.edu/catalog?utf8=✓&locale=en&search_field=all_fields&q=MSS1218_B016_I085

https://curate-test.library.emory.edu/catalog?utf8=✓&locale=en&search_field=all_fields&q=DPTEST01

devanshu-m:
@eporter23 If a source collection is present and is different from the deposit collection, then display source collection. Else display deposit collection. I apologize if that wasn't the expected behavior from this ticket. Should we create a new ticket with new a wireframe for what is expected?

eporter23:
@devanshu-m apologies if the ticket description was confusing. I think what you've implemented is fine for now and I will make sure to note the behavior. If users click through to the work page they will see both relationships, and they can also see both relationships in the dashboard views.

eporter23:
Tested in prod using some works in the Regression Test source collection since we haven't ingested the rest of Langmuir yet - works as described, thanks!

PR #1496, opened 2020-09-17
Adds logic to display source col on search results page

devanshu-m:
Connected to: #1490 

* We are adding logic to display the source col for a work on the search results page if source collection is present. If source and deposit cols are same, do nothing.


alexBLR (approved):
👍

Issue #1479, opened 2020-09-10
Display source and deposit collections in Dashboard > Works listing


eporter23:
See [wireframe item 3](https://docs.google.com/drawings/d/16I0cKv7kwZ7OeDWW6KW_xlnx2-3MYBJ3D1vrAccD0gU/edit).

When browsing works in the dashboard, the item listing should show a source collection and a deposit collection if they are distinct for a given work.

eporter23:
Discussed an alternative layout with @bwatson78 which is simpler to output:

![Screen Shot 2020-09-15 at 1.55.20 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/f3ac1461-71c5-4f2a-9447-68acbd1f439b)

lovinscari:
Devanshu will look at this after sprint review.

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1500

eporter23:
Looks good in test!

![Screen Shot 2020-10-02 at 9.46.22 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/4de3eae5-4e7b-4b20-86cf-c88cb3aa26fd)

eporter23:
Works in prod as well, thanks.

PR #1500, opened 2020-09-18
Displays Source alongside Deposit Collection in Is Part Of on Dashboard Works (#1479).

bwatson78:
`app/helpers/hyrax/collections_override_helper.rb`:
> This overrides only one helper method in `Hyrax::CollectionsHelper`, namely `render_collection_links(solr_doc)`. Lines 10 and 11 are the only changes where we are pulling source collections attached to the work, and integrating them into the `collections_list` array that the "Is Part Of:" is fed from.

Note: I couldn't get rspec to load the "Dashboard > All Works" page after associating the work with the deposit collection (e.g. `work.member_of_collections << deposit_collection`). Because of this, I have left with out of the system rspec tests.

### DIFF
--- a/app/helpers/hyrax/collections_override_helper.rb
+++ b/app/helpers/hyrax/collections_override_helper.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module CollectionsOverrideHelper
+    include Hyrax::CollectionsHelper
+
+    # rubocop:disable Rails/ContentTag
+    def render_collection_links(solr_doc)
+      collection_list = Hyrax::CollectionMemberService.run(solr_doc, controller.current_ability)
+      source_collection = collection_list.present? ? [::SolrDocument.find(collection_list.first['source_collection_id_tesim'])] : []
+      collection_list |= source_collection
+      return if collection_list.empty?
+      links = collection_list.map { |collection| link_to collection.title_or_label, hyrax.collection_path(collection.id) }
+      collection_links = []
+      links.each_with_index do |link, n|
+        collection_links << link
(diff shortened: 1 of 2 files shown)


bwatson78:
Screen Shot 2020-09-15 at 1 55 20 PM

devanshu-m:
@bwatson78 Looks good. What's the error with the system test? Can you push the tests to this branch and I can take a look if you want.

bwatson78:
It's actually not technically an error. It's just that, once I build a test scenario and associate the work to the the collection (`work.member_of_collections << deposit_collection`) and then try to visit the `dashboard/works` page, I get the system response saying the page doesn't exist. It exists all the way up to the point that I do the pushing of the deposit collection into `work.member_of_collections`.

devanshu-m:
Have you tried a `work.reload` after `work.save!` and then visiting the page? @bwatson78

bwatson78:
I did try the reload/save on the work as well as all of the collections I created--didn't change the outcome. And I also tried just simulated clicking of the Works and then All Works tab starting from the Dashboard, but still got the page missing html.

devanshu-m:
Makes me wonder if there's a small chance of this error occurring in the ui at some point?

bwatson78:
I'm just not sure it has anything to do with the helper I changed. It's probably just that I'm missing a step building the works.

devanshu-m:
@bwatson78 are you working on making any changes to this?

bwatson78:
I haven't made any changes. Per my last comment, I'm confident that my changes aren't making that error in testing.

devanshu-m (approved):

Issue #1504, opened 2020-09-21
Suppress Add to Collection button on view work pages for non-Admin users


eporter23:
The Add to Collection button is prominent on View Work pages and can be accessed by those who have edit permission to a work and can be triggered while simply viewing a work (outside the Edit form). We should suppress this from display to reduce accidental moving of works between collections and to reduce collections membership changes on large collections. 

![Screen Shot 2020-09-21 at 8.51.44 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/1d96034d-bb20-49ea-bd18-b2c0fbd2b333)

bwatson78:
@eporter23 requested that this be commented out so that it could be added back in easily.

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1514

![Screen Shot 2020-09-24 at 11.11.34 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/263d3706-85f5-4474-94ab-a2a65875c364)

eporter23:
Tested successfully with

- Viewer-level user
- Manager-level user

eporter23:
Tested with a viewer-only and manager permissions in prod. Looks good!

PR #1514, opened 2020-09-24
Suppresses Add to collection button on Works page when user non-admin (#1504).

bwatson78:
- app/views/hyrax/base/_show_actions.html.erb: adds admin requirement for Add to collection button.
- spec/system/viewing_a_work_spec.rb: tests for existence of button when user admin/non.


bwatson78:
Screen Shot 2020-09-24 at 11 11 34 AM

devanshu-m (approved):

Issue #1249, opened 2020-06-09
FileSet attached User Activity data always shows depositor name


eporter23:
When a user makes file-related edits to an object such as adding a new fileset, the User Activity shows the action with the original depositor's name even if a different person adds the FileSet.

Example - Kathryn deposited this work via bulk ingest.
I later added a fileset to the work.
The user Activity trail shows Kathryn having added the new fileset.

![Screen Shot 2020-06-09 at 4.26.35 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/a8a051c8-5ca7-477b-a888-8c45ff3d4077) 

Note: making metadata edits to a FileSet _does_ show the initiating user (when different from the depositor).

https://curate-test.library.emory.edu/concern/file_sets/948sbcc2s6-cor/

![Screen Shot 2020-09-28 at 11.48.54 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/9fc75caf-6db7-42c5-ad38-cb0a51a7bc85)

eporter23:
This appears to have been fixed in another ticket, so this is probably a duplicate.

eporter23:
I thought this had been resolved, but I am seeing it again as of 9/15/20. 
I manually added this PDF fileset to a work, but it shows up as Kathryn having attached it:

https://curate.library.emory.edu/concern/file_sets/0171vhhmtf-cor?locale=en

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1517

@eporter23 The other two events we discussed on Slack (`:after_destroy` and `:after_revert_content`) are only called from a controller and strictly utilize the `current_user` object, so no change is required to them.

devanshu-m:
@eporter23 this ticket method definitions for attaching files to work on the backend. It would be a good idea to test ingests both manually and via the importer.

eporter23:
@bwatson78 @devanshu-m I've done some testing to attach filesets both in the form and through an import. I am seeing the correct username now being assigned in the User Activity area, but I am seeing that when I use the form, all of the filesets I create attach without being characterized. 

I also ran a small book import and 2/5 of the filesets were attached without characterization. 

While this does occasionally happen particularly with large bulk ingest, I'm concerned about the behavior with the form, because that is unusual behavior.

bwatson78:
@eporter23 I'll tinker with this more.

bwatson78:
[...]

PR #1517, opened 2020-09-29
Changes FileSet Attachment Event to reflect initiating user (#1249).

bwatson78:
- .rubocop.yml: allows Hyrax-imported spec to pass rubocop linting.
- app/actors/hyrax/actors/create_with_files_actor.rb: passes initiating user to the AttachFilesToWorkJob.
- app/jobs/attach_files_to_work_job.rb: takes in and makes the initiating user the priority for the user associated for the File Attached change event.
- spec/*: changes expectations for arguments passed through as well as makes sure that the `FileSetAttachedEventJob` gets the right user.

### DIFF
--- a/app/actors/hyrax/actors/create_with_files_actor.rb
+++ b/app/actors/hyrax/actors/create_with_files_actor.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+# [Hyrax-overwrite-v3.0.0.pre.rc1] Overwritten so that the associated change event
+#   reflects the initiating user, not the depositor.
+
+module Hyrax
+  module Actors
+    # Creates a work and attaches files to the work
+    class CreateWithFilesActor < Hyrax::Actors::AbstractActor
+      # @param [Hyrax::Actors::Environment] env
+      # @return [Boolean] true if create was successful
+      def create(env)
+        uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+        user = env&.user
+
+        validate_files(files, env) &&
+          next_actor.create(env) &&
+          attach_files(files, env.curation_concern, attributes, user)
+      end
+
+      # @param [Hyrax::Actors::Environment] env
+      # @return [Boolean] true if update was successful
+      def update(env)
+        uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+        user = env&.user
+
+        validate_files(files, env) &&
+          next_actor.update(env) &&
+          attach_files(files, env.curation_concern, attributes, user)
+      end
+
+      private
+
+        def filter_file_ids(input)
+          Array.wrap(input).select(&:present?)
+        end
+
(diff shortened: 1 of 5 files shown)


bwatson78:
No screenshots necessary.

app/actors/hyrax/actors/create_with_files_actor.rb
```diff
@@ -0,0 +1,69 @@
...
+    # Creates a work and attaches files to the work
+    class CreateWithFilesActor < Hyrax::Actors::AbstractActor
+      # @param [Hyrax::Actors::Environment] env
+      # @return [Boolean] true if create was successful
+      def create(env)
+        uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+        user = env&.user
```
devanshu-m, line 16: Can you please add a comment here that this is new in this overwritten file

app/actors/hyrax/actors/create_with_files_actor.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      # @param [Hyrax::Actors::Environment] env
+      # @return [Boolean] true if update was successful
+      def update(env)
+        uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+        user = env&.user
```
devanshu-m, line 30: Same comment here please

app/actors/hyrax/actors/create_with_files_actor.rb
```diff
@@ -0,0 +1,69 @@
...
+      def create(env)
+        uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+        user = env&.user
+
+        validate_files(files, env) &&
+          next_actor.create(env) &&
+          attach_files(files, env.curation_concern, attributes, user)
```
devanshu-m, line 20: Can you please add a comment here that user is the new arg

app/actors/hyrax/actors/create_with_files_actor.rb
```diff
@@ -0,0 +1,69 @@
...
+      def update(env)
+        uploaded_file_ids = filter_file_ids(env.attributes.delete(:uploaded_files))
+        files             = uploaded_files(uploaded_file_ids)
+        # get a current copy of attributes, to protect against future mutations
+        attributes        = env.attributes.clone
+        user = env&.user
+
+        validate_files(files, env) &&
+          next_actor.update(env) &&
+          attach_files(files, env.curation_concern, attributes, user)
```
devanshu-m, line 34: Same comment here please

devanshu-m:
@bwatson78 any comment on the argument order being different in the two files?

bwatson78:
@devanshu-m I didn't see that you requested that, but If you'd like me to add the comment, I will.

devanshu-m:
@bwatson78 I meant wouldn't it cause issues if the order of args is different in one file as compared to the other? User is being passed as the fourth arg but when called/received, it is the third arg. Will this cause an issue? I understand optional args need to go towards the end, but I think this makes it necessary even when passing them, not just calling them.

devanshu-m (approved):

Issue #1522, opened 2020-10-02
CSV import option does not show all Collections to select from


eporter23:
While testing Curate v.1.2.1 I am encountering an issue where not all of the Collections are showing up in the drop-down menu. 

I am trying to add a book to the Chester W. Topp Yellowbacks collection but it does not display in the list. There are 12 collections in test, including the 5 "real" Collections, but only 10 show up in the dropdown. The Yellowbacks collection isn't shown as an option. Is this list being truncated?

**Collections list in test (10/12 displayed ) :**

![Screen Shot 2020-10-02 at 12.37.30 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/80a2d85b-ef00-45f5-b48e-321d5622ad4b)

**Collections list in production (9/9 displayed):**

![Screen Shot 2020-10-02 at 12.38.24 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/fb3f48a4-0dd3-43d5-af77-ebc5a2cc479f)

bwatson78:
@eporter23 It seems that two of the Collections are not indexed via Solr:

`collections = ActiveFedora::SolrService.query('has_model_ssim:Collection')`
`irb(main):003:0> collections.size`
`=> 10`
`irb(main):004:0> Collection.all.size`
`Attempted to init base path `prod`, but it already exists`
`=> 12`

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1528

PR #1528, opened 2020-10-05
Diverts collections options to already created custom helper (#1522).

bwatson78:
app/views/zizia/csv_imports/_collection_selection.html.erb: utilizes already created helper to ensure that all Collections show in select options.


bwatson78:
Screenshot not necessary. There is no testing needed since this is using an already tested custom helper.

devanshu-m (approved):

Issue #1511, opened 2020-09-23
Remove "Share Your Work" button from Curate homepage


eporter23:
This option should be removed for all users because it bypasses the Collection deposit process and will create works that do not receive any Collection group permissions.

To remove (for all user levels):

- [x] Share your work button
- [x] Terms of Service link

![Screen Shot 2020-09-23 at 2.56.39 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/529cd844-cb15-416b-b42d-545277c6faf4)

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1540

![Screen Shot 2020-10-15 at 10.26.57 AM.png](https://images.zenhubusercontent.com/5e6f6c00c3241244877b4aea/fc263be3-5fc2-4951-9054-fd1cff306595)

eporter23:
looks good!

PR #1540, opened 2020-10-15
Removes two unnecessary links from the homepage (#1511).

bwatson78:
- app/views/hyrax/homepage/index.html.erb: removes logic that delivers both links.
- spec/system/viewing_homepage_spec.rb: adds test to check for removal of links.


bwatson78:
Screen Shot 2020-10-15 at 10 26 57 AM

alexBLR (approved):
👍

Issue #673, opened 2019-10-11
Show User UI: Works Created link returns no matches


eporter23:
This occurs both in our local Curate as well as Nurax, so it may be a Hyrax bug.

In the [Show Users page](https://curate-test.library.emory.edu/users), if you click on the _Works Created_ link for a user with activity noted, the query returns no results.

Also, the _Activity Tab_ on an individual user's profile displays differently formatted but functional links.
![Screen Shot 2020-08-24 at 4.18.26 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/9465ae2f-f521-402a-9ff5-7511dc773c71)

**Broken query link:**
https://curate.library.emory.edu/catalog?f%5Bgeneric_type_sim%5D%5B%5D=Work&locale=en&q=%22eporter%40emory.edu%22&search_field=depositor

![Screen Shot 2019-10-11 at 10.20.57 AM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/79d7e5ee-b39c-4c4c-98c4-a889e0eecbae)

bwatson78:
@eporter23 Hyrax is configured out of the box to reference depositor to the User's email. Our depositor fields receives `uid`, so I will customize the Hyrax behavior to reflect that.

bwatson78:
@eporter23 
> Also, the Activity Tab on an individual user's profile displays differently formatted but functional links.
So, User Activity is dynamically formatted by the Actor Stack and the formats vary by Works' and Collections' attributes created. I'm not sure what you meant by the quote above. Is there some action you wanted us to take on the Activity List of items?

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1547

No screenshots necessary.

kmichaelis:
@bwatson78 this looks good to me.

eporter23:
Working as expected, thanks!

PR #1547, opened 2020-10-20
Fixes the Collections and Works created links on the User show page (#673).

bwatson78:
- app/views/hyrax/users/_vitals.html.erb: brings in override so that links can be corrected.
- spec/system/viewing_a_user_spec.rb: tests the link corrections.


bwatson78:
No screenshots necessary.

alexBLR (approved):
👍

Issue #1680, opened 2021-09-24
After All Files Updated: Debug Hyrax v3.1.0


bwatson78:
Address all errors after we have merged all updates to our v3.1.0 branch.

bwatson78:
PRs so far:
- https://github.com/emory-libraries/dlp-curate/pull/1689
- https://github.com/emory-libraries/dlp-curate/pull/1690

eporter23:
Closing because we reverted to 3.0.2.

PR #1692, opened 2021-09-28
Reverts testing to just our customized expectations (#1680).

bwatson78:
spec/presenters/collection_presenter_spec.rb: we originally brought this file in as an overwrite to just test our customizations (described in the top comments). Since we have only altered a small portion of the presenter itself, but the original Rspec assumes a clean install with all Basic Metadata included, the original Rspec really doesn't apply to us.


devanshu-m:
Are we using behavior from the original collection_type? If we are overwriting that file and using the original behavior, it might be best to keep these tests around. What do you think? @bwatson78

bwatson78:
@devanshu-m It may be worth while, but is it worth our time now? From what I can tell, we have altered/customized the Collection item so much, that the original tests shouldn't apply. I felt like I could take just a day to alleviate all of the original expectations' failures--if we're trying to get green by today, that's not gonna do it.

devanshu-m:
It might be worth the effort. But let's keep this here for now and get everything else green except for this one file. Sound ok? @bwatson78

bwatson78:
Yeah, that's fine

PR #1734, opened 2022-01-06
Updates Curate to Hyrax v3.1.0.

bwatson78:
- .circleci/config.yml: matches the ruby and builder versions used.
- .ruby-version: points to the needed ruby version.
- Gemfile*: ups the versions.
- app/models/curate/collection_type.rb and app/views/hyrax/base/_show_actions.html.erb: change that new Rails version needs.
- config/uv/*: rewrites both files to accommodate new generator output.
- package.json: alters requirements based on target Hyrax version.
- public/assets/* and yarn.lock: automatically generated by asset pre-compiling.
- spec/services/characterization_service_spec.rb: updates new expectations.


abelemlih:
Refer to #1964

PR #1735, opened 2022-01-07
Install Python2 in CI

rotated8:
Until we can figure out our Node/Yarn/Webpacker dependencies, we need Python2 to install our node modules in CI.


bwatson78:
This code is now in `main`.

PR #1773, opened 2022-02-16
Road to Hyrax 3.3.0.

bwatson78:


rotated8:
The decrease in code coverage is acceptable to me.

spec/factories/collections.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# [Hyrax-overwrite-v3.0.0.pre.rc1]
+# [Hyrax-overwrite-v3.1.0 (This may have to change for v3.2.0)]
```
abelemlih, line 3: We need to update the overwrite to `3.3.0`. Also, is any change needed for 3.3.0?

spec/jobs/attach_files_to_work_job_spec.rb
```diff
@@ -30,6 +30,7 @@
 
   shared_examples 'a file attacher', perform_enqueued: [described_class, IngestJob] do
     it 'attaches files, copies visibility and permissions and updates the uploaded files' do
+      # The following line may have to change for Hyrax v3.3.0.
```
abelemlih, line 33: Is this comment still relevant?

spec/models/collection_spec.rb
```diff
@@ -488,6 +488,7 @@
     end
 
     context "when adding members" do
+      # The code below may need to change.
```
abelemlih, line 491: Is this comment still relevant?

spec/system/create_curate_generic_work_spec.rb
```diff
@@ -278,7 +278,7 @@
     scenario "Create Curate Work" do
       visit '/concern/curate_generic_works/new'
 
-      # If you generate more than one work uncomment these lines
+      # TODO: If you generate more than one work uncomment these lines
```
abelemlih, line 281: Is this `todo` still relevant?

abelemlih (changes requested):
@bwatson78 looks great! I added a few comments/questions. 

One thing I noticed in particular is the use of `3.0.0` and `3.1.0` in files we are overwriting, instead of `3.3.0` which is the version we are upgrading to. Could please go through all the files we are overwriting, and update them with the correct version?

abelemlih:
@bwatson78 Another thing we could do, if possible, is squash these commits into one `upgrade hyrax to 3.3.0` commit for instance, to prevent having a long list of comments in history that all pertain to this one particular change. If we cannot have one commit, could we at least just reduce the number of commits by grouping some changes together, that way the commit history is cleaner once we merge this PR.

app/helpers/hyrax/file_set_override_helper.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+module Hyrax::FileSetOverrideHelper
+  include Hyrax::FileSetHelper
+
+  ##
+  # @todo inline the "workflow restriction" into the `can?(:download)` check.
```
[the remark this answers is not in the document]
bwatson78: This was another inclusion by Samvera's contributors. Since they added the comment, I thought that it would benefit us to transfer it to our override so that didn't have to go hunting for it later.

app/jobs/characterize_job.rb
```diff
@@ -40,4 +59,11 @@ def channels(filepath)
       end
       [ch]
     end
+
+    ##
+    # @api public
+    # @return [#run]
+    def characterization_service
+      self.class.characterization_service
```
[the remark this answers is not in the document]
bwatson78: This is another case where I felt it was better to mimic the structure that Samvera created to make it easier to compare the differences using `git diff`, for example. I agree with you, I wouldn't make it a private method myself--but trying to keep things as close to the original file, I believe, has benefits.

bwatson78:
@abelemlih I believe I've addressed all your requests. Let me know if you have opinions on my responses.

abelemlih (approved):
Looks great, thanks for making all the edits!

PR #1845, opened 2022-07-12
Fixes the Pub Key error showing in the Install Python2 process.

bwatson78:


bwatson78:
@rotated8 This change fixes the error, but is the change safe for deployment?

.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ jobs:
 
             - run:
                 name: Install Python2
-                command: sudo apt-get update && sudo apt-get install -y python2
+                command: sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4EB27DB2A3B88B8B && sudo apt-get update && sudo apt-get install -y python2
```
rotated8, line 38: Does this work if it's split into two lines?
```suggestion
                command: sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4EB27DB2A3B88B8B
                command: sudo apt-get update && sudo apt-get install -y python2
```

rotated8 (changes requested):
Can we split the one line into two?

.circleci/config.yml
rotated8: Wait, no, that won't work.

rotated8, line 38:
```suggestion
                command: |
                    sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4EB27DB2A3B88B8B
                    sudo apt-get update && sudo apt-get install -y python2
```

rotated8 (changes requested):
I think I fixed the syntax. I think this should work?

.circleci/config.yml
bwatson78: I'll give it a shot.

bwatson78: Last commit should address your request.

rotated8:
@bwatson78 I think this PR is no longer necessary, correct?

bwatson78:
Right.

PR #1857, opened 2022-07-28
Swaps out non-functional prepends to straight overrides.

bwatson78:
- .rubocop.yml: skips rules for overridden files.
- app/models/concerns/hydra/access_controls/access_right.rb and app/models/concerns/hydra/access_controls/visibility.rb: incorporates our customizations from the prepends into the actual class and module.
- config/initializers/prepends.rb: removes the calls out to prepend since they don't work after bumping ruby to v2.7.4.
- config/prepends/*: removes the non-functioning modules.


rotated8:
Here's how Princeton added [custom visibility levels in Figgy](https://github.com/pulibrary/figgy/commit/9fe0f58521f718f25d6b2727bdb6dbe9577cdf97). We don't have exactly the same setup, but there are probably details of their work we could use as inspiration when we get a chance to rewrite this.

alexBLR:
@bwatson78  are we going to make changes based on  @rotated8 suggestions or move on and do it later?

bwatson78:
@rotated8 and @alexBLR I'm going to attempt to do something like Figgy's implementation.

alexBLR (approved):

Issue #1868, opened 2022-08-16
Upgrade Presenters to Hyrax v3.4.1


bwatson78:
Moving from v3.0.0.pre.rc1 to v3.4.1 requires reviewing [~2000 commits](https://github.com/samvera/hyrax/compare/v3.0.0.pre.rc1...v3.4.1).

- [x] Review all the commits between v3.0.0.pre.rc1 and v3.4.1
- [x] Identify any files in`app/presenters` that we have overridden.
- [x] Update affected files to the new Hyrax version OR remove our overwrites.

bwatson78:
No changes necessary:

- dlp-curate/app/presenters/curate/file_set_presenter.rb
- 

Only Updating Version Number:
- dlp-curate/app/presenters/hyrax/admin_set_presenter.rb
-

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1870

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1873

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/1874

PR #1873, opened 2022-08-17
Updates the affected methods of Hyrax::CollectionPresenter (#1868).

bwatson78:
Updates the methods that perfectly match the original version's functions. 

Note: Collection Branding is updated here and may be worthwhile revisiting this file if we encounter problems displaying the collections' banners.

### DIFF
--- a/app/presenters/hyrax/collection_presenter.rb
+++ b/app/presenters/hyrax/collection_presenter.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+# [Hyrax-overwrite-v3.4.1]
 
 module Hyrax
   class CollectionPresenter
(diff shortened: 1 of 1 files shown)


abelemlih (approved):

app/presenters/hyrax/collection_presenter.rb
```diff
@@ -154,25 +172,21 @@ def show_path
       Hyrax::Engine.routes.url_helpers.dashboard_collection_path(id, locale: I18n.locale)
     end
 
+    ##
+    # @return [#to_s, nil] a download path for the banner file
```
abelemlih, line 176: If we're using YARD, could you please replace `[#to_s, nil]` with `[String, nil]`. I am assuming this method either returns a string or nil.

abelemlih (approved):

Issue #1921, opened 2022-08-25
Correct Failing rSpec Tests After Updating All Files


bwatson78:

bwatson78:
PRs made:
https://github.com/emory-libraries/dlp-curate/pull/1923
https://github.com/emory-libraries/dlp-curate/pull/1924
https://github.com/emory-libraries/dlp-curate/pull/1926

PR #1931, opened 2022-08-30
Fixes the spec revolving around Hyrax::FileSetsController (#1921).

bwatson78:
Summary: This is a case where, technically, this should have failed before now, but gracefully passed. The failing test ` Hyrax::FileSetsController when signed in #update when updating the attached file direct upload spawns a ContentNewVersionEventJob` was expecting an `uploader` attribute from `UploadedFile` that was an alias of the one type of `UploadedFileUploader` that Hyrax supports out of the box. We have overridden that class to allow multiple types of file to be saved to the object. Because we don't need `uploader`(we assign the path argument further down the chain), we simply commented it out and the tests were happy. 
Now, this test (`spec/controllers/hyrax/file_sets_controller_spec.rb` L# 113) is checking for the presence of any method named `uploader`, so I've created one that returns the first uploader it finds with presence. I also had to stub the heck out of the test to make it happy, even though we're only testing to see if a Job kicks off.

- app/models/hyrax/uploaded_file.rb: added method to replace alias found in original method.
- spec/controllers/hyrax/file_sets_controller_spec.rb: 
  - reworked the test on L# 113 to allow for the processing to go through.
  - updated the rest of the rSpec to match the refactoring that Samvera has done.

### DIFF
--- a/spec/controllers/hyrax/file_sets_controller_spec.rb
+++ b/spec/controllers/hyrax/file_sets_controller_spec.rb
@@ -415,5 +415,19 @@ RSpec.describe Hyrax::FileSetsController, :clean do
                                       'You need to sign in or sign up before continuing.')
       end
     end
+
+    describe '#update' do
+      it 'requires login' do
+        post :update, params: { id: public_file_set }
+        expect(response).to fail_redirect_and_flash(main_app.new_user_session_path, 'You need to sign in or sign up before continuing.')
+      end
+    end
+
+    describe '#destroy' do
+      it 'requires login' do
+        delete :destroy, params: { id: public_file_set }
+        expect(response).to fail_redirect_and_flash(main_app.new_user_session_path, 'You need to sign in or sign up before continuing.')
+      end
+    end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/controllers/hyrax/file_sets_controller_spec.rb
```diff
@@ -312,18 +370,17 @@
...
           get :show, params: { id: public_file_set }
-          expect(response).to be_success
+
+          expect(response).to be_successful
         end
       end
     end
   end
 
   context 'when not signed in' do
```
abelemlih, line 380: @bwatson78 can you add tests here for other actions, including `#delete` and `#update`, to ensure the app responds correctly when an unauthenticated user attempts those actions?

abelemlih (changes requested):

spec/controllers/hyrax/file_sets_controller_spec.rb
bwatson78: https://github.com/emory-libraries/dlp-curate/pull/1931/commits/d16e9f8612a75e2ea1dcd4770414e9c2d18a07d3 addresses your request.

abelemlih (approved):

Issue #918, opened 2020-01-28
Enhance ability to select Collection thumbnail


eporter23:
We have restored the ability to assign a thumbnail to a collection, but the current Hyrax UI option does not allow you to select a specific work - it appears to arbitrarily pick a work and its filesets to choose from.

Ideally, a user could assign a specific FileSet ID and/or perform a search to select a specific work within the collection, and then choose a FileSet within the specified work.

Because our FileSets use generic naming conventions such as PDF for volume, page 1, image 1, etc, it is impossible to know from the Hyrax selection UI which work is being presented.

Ideas/Options:
* A user can enter the ID for a specific fileset in the Edit Collection form (user goes to FileSet page, copies ID from URL)
* OR - the Collection form is extended so that a user can search for a work by its Title or ID within the Collection, select it, and then choose the appropriate fileset from a list

tmiles2:
@alexBLR can you provide an update on this ticket? where it stands, the remaining work.

PR #1972, opened 2022-09-19
Fixes edit_collection_spec

alexBLR:


bwatson78:
@alexBLR Are you sure this is the expected display value? Your work [here](https://github.com/emory-libraries/dlp-curate/issues/918) brought this from the Thumbnail Title to the id that this rspec expected.

PR #2016, opened 2022-10-26
Fix rubocop issues

abelemlih:


Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'coffee-rails', '~> 4.2'
 gem 'devise'
 gem 'devise-guests', '~> 0.6'
 gem 'dotenv-rails'
+gem 'dry-monads', '~> 1.4.0'
```
abelemlih, line 20: `dry-monads v1.5.0` causes issues with Rspec, which is why I set the version to `'~> 1.4.0'`. The error triggered during testing was:

```
An error occurred while loading ./spec/system/importer_preprocessed_langmuir_spec.rb.
Failure/Error: require File.expand_path('../../config/environment', __FILE__)

NameError:
  uninitialized constant Dry::Monads::Result::Transformer
```

.rubocop.yml
```diff
@@ -103,6 +103,8 @@ Metrics/ParameterLists:
         - app/importers/yellowback_preprocessor.rb
         - lib/importing_modules/file_set_methods.rb
         - app/jobs/manifest_persistence_job.rb
+        - app/jobs/characterize_job.rb
+        - config/initializers/collection_member_search_service_overrides.rb
```
abelemlih, line 107: `collection_member_search_service_overrides` in Hyrax v3.4.2 disables the same rule in Rubocop

spec/controllers/hyrax/file_sets_controller_spec.rb
```diff
@@ -112,15 +112,12 @@
 
       context "when updating the attached file direct upload" do
         let(:actor) { double }
-        let(:uploader) { double }
-        let(:fake_file) { double }
+        let(:pmf) { File.open(fixture_path + '/world.png') }
+        let(:uploaded_file) { Hyrax::UploadedFile.new(user: user, file_set_uri: file_set.uri, preservation_master_file: pmf) }
```
abelemlih, line 116: Rspec no longer allows the use of `allow_any_instance_of`.

The core issue in this test was that the uploaded file was created without any `preservation_master_file `, or any other file that could be returned by the `uploader` method in `app/models/hyrax/uploaded_file.rb`:

```
 def uploader
      files = [service_file, preservation_master_file, intermediate_file, extracted_text, transcript, collection_banner]
      files.find(&:present?)
end
```

Gemfile
```diff
@@ -17,7 +17,9 @@ gem 'coffee-rails', '~> 4.2'
 gem 'devise'
 gem 'devise-guests', '~> 0.6'
 gem 'dotenv-rails'
-gem 'edtf-humanize'
+gem 'dry-monads', '~> 1.4.0'
+gem 'edtf', '~> 3.0.0'
```
abelemlih, line 21: I am specifying the versions here because upgrading to the latest version of the `edtf` gem causes the following test failure in our system:

```
Failure/Error: expect(service.human_readable_date('2019-XX-07')).to eq('2019-XX-07')

  expected: "2019-XX-07"
       got: "January 7, 2019"

  (compared using ==)
./spec/services/date_service_spec.rb:66:in `block (3 levels) in <top (required)>'
```

I am keeping the same versions we currently use in `main`. If we decide to upgrade these gems, we can do so in a separate PR once the upgrade branch is merged.

bwatson78 (approved):

PR #2072, opened 2022-12-22
Fix FileSetAttachedEventJob related bug

abelemlih:
Fixes the following [issue](https://app.honeybadger.io/projects/66999/faults/91028530) reported on Honeybadger.

I noticed that across Hyrax, whenever ordered members are assigned to a work, it is often done within the constrains of an acquired lock for the work. See `app/actors/hyrax/actors/ordered_members_actor.rb #attach_ordered_members_to_work` and `app/actors/hyrax/actors/file_set_actor.rb #attach_to_work`.

In this PR, I restored the assignment of ordered members to the work in `attach_files_to_work_job.rb` to also be within the same constraints, reverting the implementation of method `#attach_to_af_work` to the implementation in Hyrax v3.4.2.

### DIFF
--- a/spec/actors/hyrax/actors/file_set_actor_spec.rb
+++ b/spec/actors/hyrax/actors/file_set_actor_spec.rb
@@ -66,7 +66,7 @@ RSpec.describe Hyrax::Actors::FileSetActor, :clean do
         # Confirm that embargo/lease are not set.
         expect(file_set_subject).not_to be_under_embargo
         expect(file_set_subject).not_to be_active_lease
-        expect(file_set_subject.visibility).to eq 'restricted'
+        expect(file_set_subject.visibility).to eq 'open'
       end
     end
   end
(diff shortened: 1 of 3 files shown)


spec/actors/hyrax/actors/file_set_actor_spec.rb
```diff
@@ -66,7 +66,7 @@
         # Confirm that embargo/lease are not set.
         expect(file_set_subject).not_to be_under_embargo
         expect(file_set_subject).not_to be_active_lease
-        expect(file_set_subject.visibility).to eq 'restricted'
+        expect(file_set_subject.visibility).to eq 'open'
```
abelemlih, line 69: @bwatson78 By running the test manually, I also realized here that our implementation of how visibility is assigned to a file_set is not working properly. In this spec, we expect `visiblity` to be `restricted`, even though the work's visibility is `open` and the `assign_visibility?` method in the actor returns `false`.  The correct visibility should be `open` in this case. By not adding the file_set to the ordered members in our previous implementation, we never actually saved the correct visibility assigned.

bwatson78 (changes requested):

abelemlih:
Issue revolved by https://github.com/emory-libraries/dlp-curate/pull/2073

Issue #2017, opened 2022-10-27
Investigate exporting with Bulkrax


eporter23:
_As a Repository Administrator, I want to export metadata and files from the repository, so that I can assist with patron requests for full-object downloads_

_As a Metadata Specialist, I want to export metadata for ingested works, so that I can make bulk edits to a spreadsheet for re-importing back into the repository_

For this epic, we want to test Bulkrax's export capabilities for the following

- Exporting metadata only as a CSV for selected works or collections
- Exporting works' metadata and files as a BagIt or ZIP file

*Acceptance criteria*

- [x] 1 For metadata exports, the resulting CSV file should match our importer heading columns to assist with re-imports
- [x] 2 Only Repository Administrators can see the Exporters interface
- [x] For file exports, the file must be able to be re-associated with its parent work either by its filename or other metadata
- [x] 3 The exporter UI should include a Title column like we have added for the Importers interface to assist with troubleshooting
- [x] 4 Any populated Preservation Workflow metadata for works should be included in the exported CSV output as they are formatted in CSVs for input, e.g. with columns as follows. Note that these are the most commonly populated, but this pattern extends to Decommission and Deletion entries. See tickets #791 and #1557 for prior work related to bulk import of preservation workflow data.
-- Ingest.workflow_notes	
-- Ingest.workflow_rights_basis	
-- Ingest.workflow_rights_basis_date	
-- Ingest.rights_basis_reviewer	
-- Ingest.workflow_rights_basis_note
-- Ingest.workflow_rights_basis_uri
-- Accession.workflow_rights_basis	
-- Accession.workflow_rights_basis_date	
-- Accession.workflow_rights_basis_reviewer	
-- Accession.workflow_rights_basis_note
-- Accession.workflow_rights_basis_uri
- [x] 5 A user should be able to specify a set of works/collections/filesets to export by their IDs in addition to the default Bulkrax selection options, which are not useful to many of our local needs. See [wireframe for possible implementation idea](https://docs.google.com/drawings/d/1Uk4nIU6XRIjB_-k84ZFwYxv2RPVv9RURWz49x2JBb-E/edit). The export should contain the children of the IDs specified (e.g. if a work is specified, its fileset entries should also be exported). 
- [x] 6 When fileset entries are exported, the creator column should be left blank. Currently Bulkrax assigns the depositor's user id as the creator, but this is confusing if the metadata and/or files are being shared with a patron requested the information.

eporter23:
[...]

PR #2078, opened 2023-01-04
Blocks the value of FileSet's creator from being displayed.

bwatson78:


bwatson78:
Addresses task in #2017

abelemlih (approved):

PR #2102, opened 2023-02-02
Revisions to Bulkrax v5 behaviour.

bwatson78:
- app/models/ability.rb: new permissions were undesired.
- config/initializers/bulkrax.rb: because the setter method and the "get" methods are different for this one file type, it is necessary to convert the filetype when exporting.
- spec/*: this is reverting the changes that were needed to test the new permissions.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -6,11 +6,11 @@ class Ability
   self.ability_logic += [:everyone_can_create_curation_concerns]
 
   def can_import_works?
-    can_create_any_work?
+    can? :read, :admin_dashboard
   end
 
   def can_export_works?
-    can_create_any_work?
+    can? :read, :admin_dashboard
   end
 
   # Define any customized permissions here.
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -6,11 +6,11 @@ class Ability
   self.ability_logic += [:everyone_can_create_curation_concerns]
 
   def can_import_works?
-    can? :read, :admin_dashboard
+    admin?
   end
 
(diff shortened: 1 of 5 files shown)


app/models/ability.rb
```diff
@@ -6,11 +6,11 @@ class Ability
   self.ability_logic += [:everyone_can_create_curation_concerns]
 
   def can_import_works?
-    can_create_any_work?
+    can? :read, :admin_dashboard
```
abelemlih, line 9: @bwatson78 looking at the `ability.rb` in Hyrax `v3.4.2`:

```
def admin_permissions
      return unless admin?
      # TODO: deprecate this. We no longer have a dashboard just for admins
      can :read, :admin_dashboard
      alias_action :edit, to: :update
      alias_action :show, to: :read
      alias_action :discover, to: :read
      can :update, :appearance
      can :manage, String # The identifier of a work or FileSet
      can :manage, curation_concerns_models
      can :manage, Sipity::WorkflowResponsibility
      can :manage, :collection_types
    end
```

It looks like the permission `can? :read, :admin_dashboard` should be deprecated. I was wondering if we could just check if the user is an admin in general? In case there are specific permissions an admin user must have, ideally we add them separately to our custom `ability.rb`.

abelemlih (changes requested):

app/models/ability.rb
bwatson78: That's cool, was just carrying over old code. Making changes.

abelemlih (approved):

PR #2104, opened 2023-02-07
Corrects the Google Tag script placement.

bwatson78:


rotated8:
I would have assumed we could use the [Hyrax setup for analytics](https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#analytics-and-usage-statistics). Does that not work for us?

bwatson78:
@rotated8 I will attempt the [ENV approach](https://github.com/samvera/hyrax/wiki/Hyrax-Management-Guide#analytics-and-usage-statistics) per our conversation.

bwatson78:
The `ENV` approach 500s when trying to load any dashboard page is all of the necessary Google variables are not set. Trying the `config` approach now.

bwatson78:
The same error occurs when only the config variables are set. Turning on Google Analytics in Hyrax' code means we want to deliver the stats in our own app's UI, no way around that.

bwatson78:
After discussing the full Google Analytics integration requirements, @eporter23 felt it would be better to only do the tagging script and forgo the whole integration. That makes this ticket ready for review.

alexBLR (approved):

Issue #2168, opened 2023-06-22
Populate from ASpace button - select list is not populated in production


eporter23:
When deploying the latest Curate version to production, the ASpace integration is visible in the New Collection form, but the popup window does not populate any values in the initial repository select list menu.

![Screen Shot 2023-06-22 at 2.05.50 PM.png](https://images.zenhubusercontent.com/5b6de3fda32a4f476097b0be/1ae240da-7785-464f-a810-7b8104ccc2a6)

eporter23:
@abelemlih this is now in production, but do you know if additional work is needed for this integration or is this a bug?

bwatson78:
This was due to two factors: 

1. The needed RAILS_ENVs were not in place. They now live in `.env.production`.
2. The ip address 52.207.83.54 was not whitelisted to communicate with our production ArchivesSpace. That has been corrected as well.

I have observed that the ArchivesSpace collections show in that pulldown's options.

eporter23:
@kmichaelis this should be ready for testing, if you have a new collection to setup?

kmichaelis:
The list previously appeared populated but no longer does.

bwatson78:
PR made: https://github.com/emory-libraries/dlp-curate/pull/2230

eporter23:
@kmichaelis and @bwatson78 I created a new collection in both arch and test via the Import from ASpace option. @kmichaelis do you want to/have you already done further testing, or is this okay to go to production?

bwatson78:
After working without error in Test and Arch, choosing to create a new collection produces a 500 with the following error:
`ArchivesSpace::ConnectionError (API client login failed as user [curate_api], check username and password are correct)`

I asked @erussey if this could be a whitelisting issue, she agreed and advised me to email as@lyrasis.org for further assistance. I will add a comment when I receive word back from them.

bwatson78:
<img width="1337" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMnUxQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7a1da231868b0f0d9adf8ed13240f425ed30b941/Screenshot_20240110_101752.png" alt="Screenshot_20240110_101752.png" />

bwatson78:
[...]

PR #2253, opened 2024-02-15
Revert "Moves the loading of Aspace repositories to the controller action (#2168). (#2230)"

bwatson78:
This reverts commit c57dfd0460e9fe2b74c5079c2e9abf236e0066ea.


bwatson78:
@abelemlih Since we've determined that our code isn't at fault and Lyrasis' API seems to stop accepting our password periodically, your AJAX code is preferred over a controller action so that we can load the page before running into any API tomfoolery.

bwatson78:
@abelemlih Per @rotated8, I'll be making another PR to provider a default option that reports the API isn't communicating correctly.

abelemlih (approved):

PR #2346, opened 2025-07-21
Updates UniversalViewer.

bwatson78:


bwatson78:
Note: setting the ENV `NODE_OPTIONS=--openssl-legacy-provider` is necessary for deploy.

